PR #6, opened 2024-07-01
:hammer_and_wrench: Automatically publish releases to NPM

miguelangelmorenochacon:
Any time a tag is pushed to this repository the CI will pass and a new package will be published in NPM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,7 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          cat ~/.npmrc
           yarn config set npmPublishRegistry "https://npm.pkg.github.com"
           yarn config set npmAuthToken "${NODE_AUTH_TOKEN}"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Create the new package version
         if: startsWith(github.ref, 'refs/tags/')
         env:
-          RELEASE_TAG: ${{ steps.check-tag-format.outputs.version }}
+          RELEASE_TAG: ${{ steps.check-tag-format.outputs.full_without_prefix }}
         run: |
           yarn version --no-git-tag-version --new-version "${RELEASE_TAG}"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,4 +94,4 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/')
         env:
           NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: npm publish --tag ${{steps.get-release-type.outputs.result}}
+        run: yarn npm publish --tag ${{steps.get-release-type.outputs.result}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     # FIXME:
     # branches: [main]
+    branches: [feature/sc-420243/ci-configuration]
     tags: [v*]
 
   # pull_request:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 
+env:
+  YARN_VERSION: 4.3.1
+
 on:
   push:
     # FIXME:
@@ -40,6 +43,9 @@ jobs:
 
       - name: Enable Corepack
         run: corepack enable
+
+      - name: Install yarn
+        run: npm install -g yarn@${YARN_VERSION}
 
       - name: Install
         run: yarn install
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      ###
+      # Build
+      ###
+
       # - name: Set up Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,47 +3,100 @@ name: CI
...
-    env:
-      CI: true
+      - name: Build
+        run: yarn build
 
-    steps:
-      - uses: actions/checkout@v4
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v4
+      - name: Lint
```
RuBiCK, line 48: I would run the linter before starting the build process to avoid building a version that does not pass the linter. If the tests are unit tests and don't require the build, I would also run them before the build.

miguelangelmorenochacon: Sure thing!

.github/workflows/ci.yml
```diff
@@ -3,47 +3,100 @@ name: CI
 on:
   push:
     branches: [main]
+    tags: ['*']
+
```
donmccurdy, line 7: Could we fork this into a new github workflow? So there will be two:

- `.github/workflows/ci.yml` (what's on `main` branch)
- `.github/workflows/release.yml` (what's on this branch)

Not every push or pull request will be a release, so I just want to be sure we can still run the CI separately on those changes, and only trigger the release job on a tagged push. Then we don't need the node.js version matrix in the release job, either.

.github/workflows/ci.yml
```diff
@@ -3,47 +3,100 @@ name: CI
...
-          node-version: ${{ matrix.node-version }}
-      - run: corepack enable
-      - run: yarn install
-      - run: yarn lint
+          script: |
+            const regex = /(alpha|beta)/
+            const refName = context.ref.replace('refs/tags/', '')
+            console.log(`Ref tag: ${refName}`)
+            const releaseTypeMatch = refName.match(regex)
+            if (!releaseTypeMatch) {
```
donmccurdy, line 80: Untested, but I think it would be possible to get this information from `check-version-format-action`, like...

```suggestion
            if (steps.check-tag-format.outputs.stable) {
```

... and similarly to get the prerelease tag from `steps.check-tag-format.outputs.prerelease`

.github/workflows/ci.yml
```diff
@@ -3,47 +3,100 @@ name: CI
...
+              releaseType = 'latest'
+            } else {
+              releaseType =  releaseTypeMatch[0]
+            }
```
donmccurdy, line 84: If we make a version with an unsupported prerelease tag like `v1.0.0-next.0` then I think it would be important that this either (a) cancels the release, or (b) uses `next` as the tag. Either is OK with me, but I think it will make a production release here.

.github/workflows/ci.yml
miguelangelmorenochacon: We're already doing this. The release will only be triggered when adding a new tag. Check this conditional

https://github.com/CartoDB/carto-api-client/pull/6/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR89

.github/workflows/ci.yml
miguelangelmorenochacon: That's what we're doing. If the tag is `beta.0` we're tagging it as `beta` without the `.0`

.github/workflows/ci.yml
miguelangelmorenochacon: I've checked that and it's only marked as stable in versions `>= 1.0.0`, for instance, `0.0.33` is marked as not stable

.github/workflows/ci.yml
miguelangelmorenochacon: I prefer to keep only 1 workflow in order to avoid duplicating code

donmccurdy: Ok, this works for me – we could also split it into a second job within the same workflow at some point later.

.github/workflows/ci.yml
donmccurdy: I see. Ok, perhaps after we reach 1.0.0 then.

.github/workflows/ci.yml
donmccurdy: What happens if I make a release tagged as `test` or `alhpa` (intentionally misspelled)? It looks like that will result in publishing a release with the `latest` tag instead, which would be a problem.

miguelangelmorenochacon: We can check https://github.com/nowsprinting/check-version-format-action?tab=readme-ov-file#is_stable if you prefer

donmccurdy: My only concern is that if the version is something like `v1.0.0-aphla.4` (note incorrect spelling!) this function MUST NOT return `releaseType = 'latest'`. Publishing with a misspelled tag, or throwing error, would be OK. 

It doesn't matter to me whether we use `is_stable`. It'd be OK with me if the pre- v1.0.0 releases do or do not get `latest` tags.

donmccurdy (approved):
Thank you @miguelangelmorenochacon, looks great!

Is it OK if we merge this? If changes are needed on the latest vs. prerelease tags I think we can just adjust that later.

.github/workflows/ci.yml
miguelangelmorenochacon: Check

https://github.com/CartoDB/carto-api-client/pull/6/commits/95334574b7968fa60f22714b003425cb1befc392

donmccurdy: LGTM, thank you! 👍🏻

miguelangelmorenochacon:
Merged!

PR #28, opened 2024-10-09
feat(sources): Move sources from @deck.gl/carto

donmccurdy:
Moves functions dealing with the CARTO API from `@deck.gl/carto` into `@carto/api-client`, intending that the `@deck.gl/carto` module can be more focused on visualization. For backwards compatibility, the `@deck.gl/carto` module will depend on `@carto/api-client` and re-export the functions it previously provided. The source code in this PR is minimally changed for TypeScript compiler settings and some cleanup; the unit tests required a more extensive rewrite to move from Tape to Vitest.


package.json
```diff
@@ -34,7 +34,7 @@
     "dev": "concurrently \"yarn build:watch\" \"vite --config examples/vite.config.ts --open\"",
     "test": "vitest run --typecheck",
     "test:watch": "vitest watch --typecheck",
-    "coverage": "vitest run --coverage",
+    "coverage": "vitest run --coverage.enabled --coverage.all false",
```
donmccurdy: This change is mostly unrelated, I could move it to a separate PR if we prefer. The coverage was previously showing files both in `src/` (with good coverage) and in `build/` (with 0% coverage) and this was the only way I could find to remove the latter. Perhaps there are better ways, but `--coverage.exclude` did not work.

package.json
```diff
@@ -4,7 +4,7 @@
   "repository": "github:CartoDB/carto-api-client",
   "author": "Don McCurdy <donmccurdy@carto.com>",
   "packageManager": "yarn@4.3.1",
-  "version": "0.3.1",
+  "version": "0.4.0-alpha.3",
```
donmccurdy: Publishing alpha versions for testing in `@deck.gl/carto`.

package.json
```diff
@@ -98,5 +98,6 @@
     "vite": "^5.2.10",
     "vitest": "1.6.0",
     "vue": "^3.4.27"
-  }
+  },
+  "stableVersion": "0.3.1"
```
donmccurdy, line 102: Yarn seems to do this automatically when we publish an alpha, so that it knows what the most recent stable version was if you decide to publish another? Seems harmless, and will likely disappear on the next stable version.

package.json
donmccurdy: Moving this change into:

- https://github.com/CartoDB/carto-api-client/pull/29

donmccurdy:
@padawannn I've reverted the coverage change in this PR and moved that into https://github.com/CartoDB/carto-api-client/pull/29 with more explanation!

package.json
```diff
@@ -62,6 +61,7 @@
   },
   "devDependencies": {
     "@deck.gl/aggregation-layers": "^9.0.30",
+    "@deck.gl/carto": "^9.0.30",
```
zbigg, line 64: Hmm, i'm wondering.

Shouldn't we _remove_ this dependency?

It seems like now we would have circular dep ... even if in devDepencies

donmccurdy: The dev dependency is only here (after this PR) because we need it in examples...

https://github.com/CartoDB/carto-api-client/blob/main/examples/01-pure-js/pure-js.ts

... and it's a bit hard to test the layer sources without that. So it won't be a "circular dependency" for users of either package, I think that should be OK?

PR #34, opened 2024-12-02
Spatial index widgets

juandjara:
related story: https://app.shortcut.com/cartoteam/story/454268/adapt-carto-api-client-to-be-able-to-work-with-dynamic-spatial-index


src/sources/h3-query-source.ts
```diff
@@ -30,7 +33,7 @@ type UrlParameters = {
 };
 
 export const h3QuerySource = async function (
-  options: H3QuerySourceOptions
+  options: Omit<H3QuerySourceOptions, 'spatialDataType'>
```
donmccurdy: Primarily `H3QuerySourceOptions` represents the public-facing option types of the `h3QuerySource` function. Since users would never pass the 'spatialDataType' prop here, I think we can remove it from the 'H3QuerySourceOptions' definition.

(And similar for other sources)

src/sources/h3-query-source.ts
```diff
@@ -59,7 +62,12 @@ export const h3QuerySource = async function (
   return baseSource<UrlParameters>('query', options, urlParameters).then(
     (result) => ({
       ...(result as TilejsonResult),
-      widgetSource: new WidgetQuerySource(options),
+      widgetSource: new WidgetQuerySource({
+        ...options,
+        // NOTE: passing redundant spatialDataColumn here to apply the default value 'h3'
+        spatialDataColumn,
```
donmccurdy, line 68: Hm, good catch here! I'm a little nervous we'll forget to do this for future default values, maybe we should refactor how the defaults are handled. But we can do that in another PR later, your comment is a good callout for now. 👍🏻

src/spatial-index.ts
```diff
@@ -0,0 +1,125 @@
...
+    zoom: number;
+    latitude: number;
+    longitude: number;
+  };
+}) {
+  if (source.spatialDataType === 'geo') {
+    return undefined;
+  }
+
+  const currentZoom = viewState.zoom ?? 1;
```
donmccurdy: Minor — Since `viewState.zoom` is required input, I'd slightly prefer not giving it a fallback here.

src/widget-sources/widget-base-source.ts
```diff
@@ -93,14 +93,26 @@ export abstract class WidgetBaseSource<Props extends WidgetBaseSourceProps> {
   async getCategories(
     options: CategoryRequestOptions
   ): Promise<CategoryResponse> {
-    const {filterOwner, spatialFilter, abortController, ...params} = options;
+    const {
+      filterOwner,
+      spatialFilter,
+      spatialFiltersMode,
+      spatialFiltersResolution,
```
donmccurdy: I'm not sure yet if we _should_ do this, so no need to make the change now, but ... would it work if instead of asking the user to provide `spatialFiltersResolution`, we computed it internally from the WidgetBaseSource's own props, and a (new) viewState prop that the user would pass into `getCategories` etc.? Or is this something the user needs to directly control?

I'm a bit worried about all of the default values used inside 'getSpatialFiltersResolution', and things like `tileSize`, because it would be easy for the user to forget something. It would be nice if all the default values were populated before the 'source' is passed into the `getSpatialFiltersResolution` function, and this would make implementation in the widget web components simpler.

src/spatial-index.ts
```diff
@@ -0,0 +1,125 @@
...
+  DEFAULT_AGGREGATION_RES_LEVEL_H3,
+  DEFAULT_AGGREGATION_RES_LEVEL_QUADBIN,
+} from './constants-internal';
+import {ModelSource} from './models/model';
+import {AggregationOptions} from './sources/types';
+
+const DEFAULT_TILE_SIZE = 512;
+const QUADBIN_ZOOM_MAX_OFFSET = 4;
+
+export function getSpatialFiltersResolution({
```
donmccurdy: Does similar logic exist in C4R now, and could you share a link to it for comparison? 

I'm not sure about using `tileSize` without also referring to `tileResolution`, that seems a bit strange to me, but maybe this is just how it already works?

donmccurdy (commented):
Thanks, this is nicely structured! A few comments/questions, still working my way though the code review. :)

src/spatial-index.ts
juandjara: Yes this is almost copied verbatim from C4R. See here https://github.com/CartoDB/carto-react/blob/master/packages/react-widgets/src/models/spatialFiltersResolution.js#L43

src/spatial-index.ts
juandjara: Ok we will remove this variable then

src/widget-sources/widget-base-source.ts
juandjara: Yes, user should be able to specify a concrete `spatialFiltersResolution` I think. I'm also a bit worried about all the conditional logic inside of that function, but since is inherited legacy code, I decided not to question it very much. We can find a way to adapt it if you would like.

src/sources/h3-query-source.ts
juandjara: Okay, if we do this, where do we define the `spatialDataType` prop then?
Because the `spatialDataColumn` prop is defined in `QuerySourceOptions` and `TableSourceOptions`

juandjara: Other thing I was considering was moving this prop definition to `WidgetBaseSourceProps`

donmccurdy: The only consequence if it isn't defined explicitly (as on `main` branch now) is that the following...

```javascript
const data = await h3QuerySource({..., spatialDataType: 'quadbin'});
```

... is a runtime error instead of a compile-time error. A compile-time error would be better, I agree, but the user must be able to import and use `H3QuerySourceOptions` themselves and the required/optional fields should match what `h3QuerySource` accepts. So perhaps instead of removing the override, we can just change its (still optional) value as `spatialDataType?: 'h3'` rather than `spatialDataType: 'h3'`?

```javascript
export type H3QuerySourceOptions = SourceOptions &
  QuerySourceOptions &
  AggregationOptions &
  FilterOptions & {
    spatialDataType?: 'h3';
  };
```

src/widget-sources/widget-base-source.ts
donmccurdy: /cc @zbigg @felixpalmer does this concern make sense, and do you have any immediate preferences? 

If we're all not sure then maybe it's best to go ahead and try setting up 1-2 of the example widgets (https://github.com/CartoDB/carto-api-client/tree/main/examples/components/widgets) to test with spatial index sources.

juandjara: Okay I'll let you know when I have the examples ready

src/spatial-index.ts
donmccurdy: I see, thanks! I think at least for the `aggregationResLevel` we should make sure the user doesn't need to specify that twice (once for the source, and again for getSpatialFiltersResolution). But there's more than one way to do this, so I think it will depend on the question above.

src/sources/h3-query-source.ts
juandjara: That's perfect, I was working on a second proposal for this, but this one is easier to understand

src/widget-sources/widget-base-source.ts
felixpalmer: Bear in mind I am not familiar with this code, so feel free to disregard the following :)

Am I understanding correctly that by providing `spatialFiltersResolution` the user will be able to override what `getSpatialFiltersResolution` would otherwise provide?

`getSpatialFiltersResolution` is a complex function and especially with h3 there has been a lot of confusion about what "resolution" means and how it changes with the viewport.

As such I would vote for not exposing this parameter directly, but instead let the user specify a `spatialResolutionBias`, i.e. a value which increases/decreases the resolution that would otherwise be used. So passing `2` would give double the resolution compared to the default.

src/sources/h3-query-source.ts
juandjara: Second proposal has been merged, I'm going to mark this as resolved

src/widget-sources/widget-base-source.ts
juandjara: @felixpalmer I agree `getSpatialFiltersResolution` is a complex function and does not need to be exported

We have added more options with merging [this PR](https://github.com/CartoDB/carto-api-client/pull/35) here, where you pass a viewState (latitude, longitude and zoom) param instead of calling the function and the function gets called internally with that. But still we are allowing to receive a `spatialFiltersResolution` param

juandjara: And for the `spatialResolutionBias`, I can see how it would be applied to h3 resolution because there is a BIAS constant there, but how would we apply it to quadbins? Makes sense to apply to one and not the other or is it just a multiplication factor?

donmccurdy: I don't think there is any use case (CARTO-internal or external) that currently requires either `spatialFiltersResolution` or `spatialResolutionBias`, correct? If not then perhaps we should just wait on exposing it.

examples/components/widgets/category-widget.ts
```diff
@@ -57,6 +57,13 @@ export class CategoryWidget extends BaseWidget {
         spatialFilter: this.getSpatialFilterOrViewState(),
         operation,
         column,
+        viewState: this.viewState
+          ? {
+              zoom: this.viewState.zoom,
+              latitude: this.viewState.latitude,
+              longitude: this.viewState.longitude,
+            }
+          : undefined,
```
donmccurdy: Not necessary to pluck the zoom/lat/long out explicitly I think, let's just do:

```javascript
viewState: this.viewState ?? undefined
```

src/models/model.ts
```diff
@@ -96,18 +105,30 @@ export function executeModel(props: {
...
 
   if (spatialFilters) {
     queryParams.spatialFilters = JSON.stringify(spatialFilters);
+    queryParams.spatialDataColumn = spatialDataColumn;
+    queryParams.spatialDataType = spatialDataType;
+  }
+
+  if (spatialDataType !== 'geo') {
+    if (spatialFiltersResolution > 0) {
+      queryParams.spatialFiltersResolution = String(spatialFiltersResolution);
```
donmccurdy: Hm, any idea why the string cast is needed here?

src/widget-sources/widget-base-source.ts
```diff
@@ -79,6 +81,8 @@ export abstract class WidgetBaseSource<Props extends WidgetBaseSourceProps> {
       filters: getApplicableFilters(owner, props.filters),
       filtersLogicalOperator: props.filtersLogicalOperator,
       geoColumn: props.geoColumn,
+      spatialDataType: props.spatialDataType,
+      spatialDataColumn: props.spatialDataColumn,
```
donmccurdy, line 80: Hm, I think it was a mistake to have `geoColumn` here at all – it should have been `spatialDataColumn` to match https://deck.gl/docs/api-reference/carto/data-sources. Feel free to remove the column here, or I can do that in a separate PR, whichever is convenient.

Filed a new issue for it to keep track:

- https://github.com/CartoDB/carto-api-client/issues/40

src/spatial-index.ts
```diff
@@ -0,0 +1,130 @@
...
+  source,
+  viewState,
+}: {
+  source: {
+    spatialDataType?: ModelSource['spatialDataType'];
+    dataResolution?: ModelSource['dataResolution'];
+    aggregationResLevel?: AggregationOptions['aggregationResLevel'];
+  };
```
donmccurdy: I think this part could be simplified as:

```suggestion
  source: Partial<ModelSource & AggregationOptions>;
```

src/spatial-index.ts
```diff
@@ -0,0 +1,130 @@
...
+  viewState?: {
+    zoom: number;
+    latitude: number;
+    longitude: number;
+  };
```
donmccurdy: Similarly:

```suggestion
  viewState?: ViewState;
```

src/spatial-index.ts
```diff
@@ -0,0 +1,130 @@
...
+    'viewState prop is required to compute automatic spatialFiltersResolution when using spatialFilter with spatial indexes. Either pass a `spatialFiltersResolution` prop or a `viewState` prop to avoid this error'
+  );
+
+  const dataResolution = source.dataResolution ?? Number.MAX_VALUE;
+
+  const aggregationResLevel =
+    source.aggregationResLevel ??
+    (source.spatialDataType === 'h3'
+      ? DEFAULT_AGGREGATION_RES_LEVEL_H3
+      : DEFAULT_AGGREGATION_RES_LEVEL_QUADBIN);
```
donmccurdy, line 23: Just as FYI, but does not need to be fixed here:

- https://github.com/CartoDB/carto-api-client/issues/39

src/models/model.ts
juandjara: Yes, this is due to declaring `queryParams` values as `string` in here

```
const queryParams: Record<string, string>
```

We could do a refactor here so we would not have to do and undo the `JSON.stringify` and convert to string everything that goes into `queryParams`. Redeclare this variable as `Record<string, unknown>` and only convert to string below the `isGet` conditional

examples/components/widgets/category-widget.ts
juandjara: Done 👍

src/spatial-index.ts
juandjara: Done 👍

src/spatial-index.ts
juandjara: Done 👍

src/widget-sources/widget-base-source.ts
juandjara: Ok I removed the `geoColumn` property in favor of `spatialDataColumn` since we were not using `geoColumn` in any other place

src/widget-sources/widget-base-source.ts
juandjara: Ok, using this commit https://github.com/CartoDB/carto-api-client/pull/34/commits/f85f818305e41218c9c2e597ee74e852c0977e05 to not export the `getSpatialFiltersResolution` function

src/spatial-index.ts
juandjara: Okay I think with the last implementation we can consider this conversation closed, right?
Since right now the user does not have to pass any source props to the widget model call

src/models/model.ts
juandjara: I've created a new PR here https://github.com/CartoDB/carto-api-client/pull/41 with a proposal for refactoring this

src/spatial-index.ts
donmccurdy: Yes, I think so! I still want to look at https://github.com/CartoDB/carto-api-client/issues/39 later but that shouldn't block anything here.

src/widget-sources/widget-base-source.ts
```diff
@@ -93,14 +94,35 @@ export abstract class WidgetBaseSource<Props extends WidgetBaseSourceProps> {
...
     const {column, operation, operationColumn} = params;
+    const source = this.getModelSource(filterOwner);
+
+    let spatialFiltersResolution = options.spatialFiltersResolution;
+    if (spatialFilter && !spatialFiltersResolution) {
+      spatialFiltersResolution = getSpatialFiltersResolution({
+        source,
+        viewState,
+      });
+    }
```
donmccurdy: Following https://github.com/CartoDB/carto-api-client/pull/34#discussion_r1871356107, I think we also don't want to expose the `spatialFiltersResolution` option to users yet, so we can unconditionally compute it here, if a spatial filter is used.

I'm not opposed to adding it later, but as Felix mentions there is some subtlety about how we can/should do that, and it doesn't seem essential to get this feature shipped now.

donmccurdy:
@juandjara great work here! I think all that's left after https://github.com/CartoDB/carto-api-client/pull/41 is the one other comment above, removing the `spatialFiltersResolution` from the function options and always computing it internally when a spatial filter is used.

As the next step, should we publish this as an alpha release so that external examples can be built on it? If you're ready, feel free to do that any time. The steps would:

1. Update `package.json#version` to `0.4.1-alpha.0`
2. Run `yarn postversion`
3. Manually run the release action, selecting this branch as the target, as described in [CONTRIBUTING.md](https://github.com/CartoDB/carto-api-client/blob/main/CONTRIBUTING.md#releases)

src/widget-sources/widget-base-source.ts
juandjara: Ok I'm merging this comment with some other from my team in PS to take the early return out of the `getSpatialFiltersResolution` and add it the conditional there

src/models/model.ts
```diff
@@ -79,50 +85,51 @@ export function executeModel(props: {
...
-  // API supports multiple filters, we apply it only to geoColumn
+  // API supports multiple filters, we apply it only to spatialDataColumn
   const spatialFilters = source.spatialFilter
-    ? {[geoColumn]: source.spatialFilter}
+    ? {[spatialDataColumn]: source.spatialFilter}
     : undefined;
 
   if (spatialFilters) {
-    queryParams.spatialFilters = JSON.stringify(spatialFilters);
+    queryParams.spatialFilters = spatialFilters; // JSON.stringify(spatialFilters);
```
donmccurdy: Minor cleanup -

```suggestion
    queryParams.spatialFilters = spatialFilters;
```

donmccurdy (approved):

donmccurdy (approved):

PR #35, opened 2024-12-03
Spatial index widgets second proposal

juandjara:
Following up on the conversation here: https://github.com/CartoDB/carto-api-client/pull/34

This is an alternative implementation for the feature of adding widgets to spatial index sources configuring spatial filter to work in that case

### DIFF
--- a/src/widget-sources/widget-base-source.ts
+++ b/src/widget-sources/widget-base-source.ts
@@ -16,9 +16,10 @@ import {
   TableResponse,
   TimeSeriesRequestOptions,
   TimeSeriesResponse,
+  ViewState,
 } from './types.js';
 import {FilterLogicalOperator, Filter} from '../types.js';
-import {getApplicableFilters, normalizeObjectKeys} from '../utils.js';
+import {assert, getApplicableFilters, normalizeObjectKeys} from '../utils.js';
 import {getClient} from '../client.js';
 import {ModelSource} from '../models/model.js';
 import {SourceOptions} from '../sources/index.js';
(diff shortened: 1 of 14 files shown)


src/widget-sources/widget-base-source.ts
```diff
@@ -184,19 +212,31 @@ export abstract class WidgetBaseSource<Props extends WidgetBaseSourceProps> {
...
+    if (spatialFilter && source.spatialDataType !== 'geo') {
+      assert(
+        options.spatialFiltersResolution ?? viewState,
+        'spatialFiltersResolution or viewState is required for using spatialFilter with spatial indexes'
+      );
+    }
+
+    const spatialFiltersResolution = options.spatialFiltersResolution ?? (
+      getSpatialFiltersResolution({ source, viewState: viewState! })
+    );
```
donmccurdy: Probably we can move the common validation into `getSpatialFiltersResolution` here, and call the function only if a spatial filter is being used?

donmccurdy (approved):
I think I prefer computing the spatialFilterResolution internally like this, thanks! This way the user can set the resolution explicitly if the want to, but they don't always need to have the right options to pass into a new function.

I'm not 100% sure about some of the properties moved around on `SourceOptionalOptions` here, but that's easier to review after merging on the feature branch so I'm not worried about it now.

@juandjara @srtena any other concerns or preferences?

src/widget-sources/widget-base-source.ts
juandjara: Ok, good call

juandjara: In this case the assert message would need to change to something like
"viewState prop is required to compute automatic spatialFiltersResolution when using spatialFilter with spatial indexes. Either pass a `spatialFiltersResolution` prop or a `viewState` prop to avoid this error"

PR #37, opened 2024-12-04
Improve Tilejson typings to include properties used in Carto platform

zbigg:
Improve `Tilejson` typings to include properties used in Carto platform:

* add types to raster tables
* add `raster_metadata` to `Tilejson` 
* add few other known fields to `Tilejson`

Followup from this discussion: https://github.com/CartoDB/cloud-native/pull/18605#discussion_r1868173403

### DIFF
--- a/src/sources/types.ts
+++ b/src/sources/types.ts
@@ -191,39 +191,51 @@ export interface Tilejson {
   attribution: string;
   scheme: string;
   tiles: string[];
-  properties_tiles: string[];
-  minresolution: number;
-  maxresolution: number;
   minzoom: number;
   maxzoom: number;
-  bounds: [number, number, number, number];
-  center: [number, number, number];
+  bounds: [left: number, bottom: number, right: number, top: number];
+  center: [longitute: number, latitude: number, zoom: number];
   vector_layers: VectorLayer[];
+
+  //
+  // Carto additions over standard Tilejson properties
+  //
+
+  minresolution: number;
+  maxresolution: number;
+  properties_tiles: string[];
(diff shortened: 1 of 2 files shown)


src/sources/types.ts
```diff
@@ -201,6 +201,29 @@ export interface Tilejson {
   vector_layers: VectorLayer[];
   tilestats: Tilestats;
   tileResolution?: TileResolution;
+
+  // extra for Carto
```
donmccurdy: Does 'extra' here mean that these fields are only meant for internal use by CARTO applications? I think it might be good to add `@internal` JSDoc annotations if so, like:

```javascript
 /**
  * Resolution of data in spatial-index dataset (e.g. H3, Quadbin).
  * @internal
  */
  dataresolution?: number;
```

src/sources/types.ts
donmccurdy, line 1: All of the new types here represent data coming from the API in a tilejson response, correct? Not data that we're creating on the frontend?

donmccurdy (approved):

src/sources/types.ts
zbigg: Yes, they are all found in responses from Carto Maps API

src/sources/types.ts
zbigg: I'll fix wording, but in principle these are Carto additions over [tilejson](https://github.com/mapbox/tilejson-spec) standard.

Yes, they probably should documented as internal. I'm ok with `@internal` - first time seeing it.

felixpalmer (approved):
Thanks for these :)

PR #41, opened 2024-12-05
Refactor model call query params

juandjara:
Following up on the conversation here: https://github.com/CartoDB/carto-api-client/pull/34#discussion_r1870342514 I've created this PR with a refactor proposal of this internal part, to reduce the amount of type casting

### DIFF
--- a/src/models/model.ts
+++ b/src/models/model.ts
@@ -7,7 +7,7 @@ import {
   SpatialFilter,
 } from '../types.js';
 import {$TODO} from '../types-internal.js';
-import {assert} from '../utils.js';
+import {assert, isPureObject} from '../utils.js';
 import {ModelRequestOptions, makeCall} from './common.js';
 import {ApiVersion} from '../constants.js';
 import {SpatialDataType, SpatialFilterPolyfillMode} from '../sources/types.js';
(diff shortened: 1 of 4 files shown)


src/models/model.ts
```diff
@@ -153,3 +141,15 @@ export function executeModel(props: {
     },
   });
 }
+
+function objectToURLSearchParams(object: Record<string, unknown>) {
+  const params = new URLSearchParams();
+  for (const key in object) {
+    if (typeof object[key] === 'object') {
```
donmccurdy: Minor – Let's use `isPureObject` from utils, instead of the typeof check here. Probably doesn't matter, but this way "null" falls through to the next case and we can do a null or undefined check.

donmccurdy (approved):
Much cleaner than we originally had, thank you!

src/models/model.ts
juandjara: And will the `isPureObject` check catch arrays too?
Right now, checking for `object` types we are stringifying object and arrays, and converting `null` to `"null"`.

However, we could separate these three cases in three conditionals to make it easier to read

juandjara: It would read as something like this
```
if (isPureObject(object[key])) {
  params.append(key, JSON.stringify(object[key]));
}
else if (Array.isArray(object[key])) {
  params.append(key, JSON.stringify(object[key]));
}
else if (object[key] === null) {
  params.append(key, 'null');
}
else if (object[key] !== undefined) {
  params.append(key, String(object[key]));
}
```

juandjara: Something Im not sure about is how we want to treat `null` values here

PR #48, opened 2024-12-23
Support aggregationExp for Table/Query Sources

jmgaya:


src/sources/types.ts
```diff
@@ -149,6 +149,15 @@ export type QuerySourceOptions = {
    *     ```
    */
   queryParameters?: QueryParameters;
+
+  /**
+   * Comma separated expressions that, in case of containing any, the source will be aggregated by geometry automatically.
```
donmccurdy: Minor, but since this is public API documentation - When reading "in case of containing any" it sounds to me like we're saying "if it contains `ANY_VALUE`" or something like that, which I don't think is the intent.

Do we mean something like this instead?

> Comma-separated aggregation expressions. If assigned on a vector source, source is grouped by geometry and then aggregated.

donmccurdy (approved):
One comment/question on the docs, otherwise LGTM!

src/sources/types.ts
jmgaya: Nice point! What I mean is, **if present**, the source is grouped by geometry, but I think your sentence is clearer, so let me adapt it 👍

donmccurdy: Looks good, thanks!

donmccurdy (approved):

PR #50, opened 2024-12-31
feat(tilesets): Support widget calculations on tilesets

donmccurdy:
Changes:

- Migrate local calculations from C4R
- Port local calculations to TypeScript
  - In some cases the types were not obvious, particularly required vs. optional parameters and overloaded types (see: SortColumns). I've used strict TS types where possible, and generally tried to make parameters required for internal functions so that defaults are applied only once at top-level calls.
- Migrate unit tests from C4R
- Add WidgetTilesetSource class
- Refactor WidgetSource class hierarchy:
  - WidgetSource
    - WidgetRemoteSource
      - WidgetTableSource
      - WidgetQuerySource
    - WidgetTilesetSource
- Add tileset example


package.json
```diff
@@ -8,7 +8,7 @@
   "homepage": "https://github.com/CartoDB/carto-api-client#readme",
   "author": "Don McCurdy <donmccurdy@carto.com>",
   "packageManager": "yarn@4.3.1",
-  "version": "0.4.5",
+  "version": "0.5.0-alpha.2",
```
padawannn, line 11: Don't forget to undo this change before merge

src/filters/tileFeatures.ts
```diff
@@ -0,0 +1,56 @@
...
+  tiles,
+  spatialFilter,
+  uniqueIdProperty,
+  tileFormat,
+  spatialDataColumn = DEFAULT_GEO_COLUMN,
+  spatialDataType,
+  options = {},
+}: TileFeatures): FeatureData[] {
+  // TODO(cleanup): Is an empty response the expected result when spatialFilter
+  // is omitted? Why not make the parameter required, or return the full input?
```
padawannn: Agree make it required, please

src/filters/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './Filter.js';
```
padawannn, line 1: Should we export the `ts` files instead of `js` files? I mean
```
export * from './Filter.ts';
```
instead of
```
export * from './Filter.js';
```

I've notice other exports like this in other places like https://github.com/CartoDB/carto-api-client/blob/main/src/index.ts#L1

src/index.ts
```diff
@@ -18,3 +18,9 @@ export {
 } from './api/index.js';
 
 export {_getHexagonResolution} from './spatial-index.js';
+
+// For unit testing only.
+export * from './filters/index.js';
```
padawannn, line 23: Same comment, should not be 
```
export * from './filters/index.ts'
```

src/operations/groupBy.ts
```diff
@@ -0,0 +1,59 @@
+import {aggregationFunctions, aggregate} from './aggregation.js';
```
padawannn, line 1: import from `./aggregation.ts`

padawannn (changes requested):
Great job @donmccurdy. Some small comments, can you take a look?

package.json
donmccurdy: This is safe to merge – if we later do `yarn version patch`, yarn will look at the `package.json#stableVersion` entry below and automatically update to 0.4.6 and not 0.5.1, so we are able to increment either the 'alpha' or the 'latest' release safely.

If there are any other worries though, I'm happy to remove it!

src/filters/index.ts
donmccurdy: We can't do `.ts` I think, the build process would have to rename the imports in `.d.ts` files to `.js` at build time, and TypeScript has for a very long time been against allowing that:

- https://github.com/microsoft/TypeScript/issues/49462

The filenames are supposed to be paths that can be resolved _after_ TypeScript compilation, so we are stuck with either `./Filter` or `./Filter.js`. Currently the codebase has a mix of both ... so that is not great, I would like to clean it up to be consistent, I'll do that in a new PR. 😅

While I'm happy with either in an app like Builder, I think that for libraries we may need to use `.js`, because that's what should appear in the published `.d.ts` files. It's required for customers using certain configurations, particularly `moduleResolution: 'nodenext'`. I've had this feedback in some of my personal repositories in the past, discussion here:

- https://github.com/donmccurdy/glTF-Transform/pull/827

src/index.ts
donmccurdy: See above – I think it's correct but will start a new PR to make this consistent at least!

src/operations/groupBy.ts
donmccurdy: See above – I think it's correct but will start a new PR to make this consistent at least!

src/filters/index.ts
donmccurdy: For now I've opened a new issue to track this:

- https://github.com/CartoDB/carto-api-client/issues/99

I'll wait to make the changes until after this PR and the raster PR are merged, so we don't have a bunch of merge conflicts.

package.json
padawannn: ok, thx!

src/filters/index.ts
padawannn: @donmccurdy, why don't we export the file without the extension as Deck.gl does? https://github.com/visgl/deck.gl/blob/master/modules/core/src/index.ts#L90 I mean

```
export * from './Filter';
```

padawannn (approved):
Thank you for your comments.  I've approved the PR but please check this comment https://github.com/CartoDB/carto-api-client/pull/50#discussion_r1948660036

src/filters/index.ts
donmccurdy: So deck.gl is using `... from './Foo'` in source files, but they've written a special build step that adds the `.js` extension to the generated `.d.ts` files:

https://github.com/uber-web/ocular/blob/6d84ec50d8c80dc2cf361e818e9edb35f597042a/modules/dev-tools/src/ts-plugins/ts-transform-append-extension/index.ts

Added in:

- https://github.com/uber-web/ocular/pull/446

Admittedly putting the `.js` extension in source code is weird/ugly... but without customizing the build process I don't know of another way to solve that problem for customers. Another option would be to bundle the `.d.ts` files into one file, so there are no imports in the output, but that will complicate the build step and I haven't found a build setup that I trust for that yet. 😕

PR #59, opened 2025-01-07
chore(examples): Remove framework-specific examples

donmccurdy:
Removes framework-specific examples from the repository. The dependencies (and necessary dependency updates) required to include three frameworks here are too much. We have framework-specific examples elsewhere in:

- https://github.com/CartoDB/carto-app-templates
- https://github.com/CartoDB/deck.gl-examples/


jmgaya (approved):
Amazing job and clean-up 💪 

I absolutely agree that repeating templates between repositories feels wrong, as they may get outdated easily, but instead, keeping them all together elsewhere feels much more natural and easier to maintain 😄

padawannn (dismissed):
I agree with the changes but I would like to validate with @srtena first

srtena:
Yes, no problem. I also think they aren't necessary and we should focus on the app-templates. Thanks team :)

PR #60, opened 2025-01-07
chore(build): Switch from microbundle to vite lib mode

donmccurdy:
Work in progress, will need a bit more testing after https://github.com/CartoDB/carto-api-client/pull/59.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     ".": {
       "types": "./build/index.d.ts",
       "require": "./build/api-client.cjs",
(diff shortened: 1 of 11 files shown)


donmccurdy:
Never mind, the configuration is too finicky and I haven't been able to get sourcemaps and code coverage working with Vite's Library Mode. There's an update to Microbundle already on the way, so let's just pin Rollup in `package.json#resolutions` for now. Opening a new PR:

- https://github.com/CartoDB/carto-api-client/pull/61

PR #61, opened 2025-01-08
chore(deps): Pin Rollup v4

donmccurdy:
Pins rollup v4 to resolutions. Previously Vite was using v4 (examples, local testing) but microbundle was using v2 (library builds). I tried switching over entirely to Vite, using Vite's [Library Mode](https://vite.dev/guide/build#library-mode) for the library builds (https://github.com/CartoDB/carto-api-client/pull/60) but — as awesome as Vite is for building applications! — it seems really finicky for building libraries, I needed a lot of unexpected configuration just to get things working, and even then I couldn't get the sourcemaps or code coverage working.

It looks like Microbundle is already working on an update for Rollup, so I'd rather keep using that for now. We can remove the pinned resolution after the next Microbundle update, I'm tracking these issues:

- https://github.com/developit/microbundle/pull/1081
- https://github.com/developit/microbundle/pull/1082
- https://github.com/developit/microbundle/pull/1083

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "@luma.gl/core": "^9.0.27",
     "@luma.gl/engine": "^9.0.27",
     "@types/json-schema": "^7.0.15",
+    "@types/react": "^18.3.18",
     "@types/semver": "^7.5.8",
     "@vitest/coverage-istanbul": "^1.6.0",
     "@webcomponents/webcomponentsjs": "^2.8.0",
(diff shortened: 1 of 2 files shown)


donmccurdy:
Aside - I'm also curious about where Vite is going with [Rolldown](https://rolldown.rs/). Once that's stable, Vite will stop using Rollup and Esbuild, and just use Rollup – maybe the configuration will be simpler then, it might be a good time to revisit Vite Library Mode.

donmccurdy:
Small additional fix here is adding `@types/react`, we don't use React in this repo really, but the types are needed by the web component examples for React integration, without them we get warnings in build logs. I removed those types prematurely in #59.

padawannn (approved):

PR #85, opened 2025-01-29
feat: Export getHexagonResolution helper

donmccurdy:
This is subjective (there is no "right" choice of hexagon resolution for a given map viewport) but we do want to be consistent and it seems that we need this function in multiple places, so I'm including the export here.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,4 +17,4 @@ export {
   requestWithParameters,
 } from './api/index.js';
 
-export {getHexagonResolution} from './spatial-index.js';
+export {_getHexagonResolution} from './spatial-index.js';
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -16,3 +16,5 @@ export {
   query,
   requestWithParameters,
 } from './api/index.js';
+
+export {getHexagonResolution} from './spatial-index.js';
```
zbigg: In deck/c4r. we've got this convention of prefixing internal apis with `_` maybe we should carry this over? 🤷 
I have no strong opinion on it.

zbigg (approved):

src/index.ts
donmccurdy: I like the idea — done thanks! I've been using the [`@internal`](https://tsdoc.org/pages/tags/internal/) TSDoc tag in some cases but a prefix would be more visible to users.

PR #87, opened 2025-01-31
feat(rasters): Support widget calculations on rasters

donmccurdy:
In large part, widgets for raster datasources reuse the same code implemented for quadbin tilesets (https://github.com/CartoDB/carto-api-client/pull/50). To accomplish this, we pre-process each raster tile, creating a 'FeatureData' key/value object for each pixel in the raster, if not excluded by the spatial filter. Creating "an object per pixel" is far from ideal, and keeping a binary representation throughout widget calculations would greatly improve performance. However, I think it's better to keep the code aligned with quadbin tilesets right now — ideally any performance optimizations would benefit both. Use of web workers should prevent calculations from dragging down the web application, even if the calculations themselves are inefficient.

Changes:

- Add WidgetRasterSource class
- Add `tileFeaturesRaster()` function, doing the heavy lifting of extracting feature data for calculations from rasters
- Add raster example
- Minor changes to example widgets, making it easier to set category colors that match pixel colors for debugging

screenshot

### DIFF
--- a/src/filters/tileFeaturesRaster.ts
+++ b/src/filters/tileFeaturesRaster.ts
@@ -64,8 +64,9 @@ export function tileFeaturesRaster({
 
       for (const band in tile.data.cells.numericProps) {
         const value = tile.data.cells.numericProps[band].value[i];
-        // TODO(cleanup): nodata should be a number, not a string.
-        if (Number(bandMetadataByName[band].nodata) !== value) {
+        const bandMetadata = bandMetadataByName[band];
+
+        if (isValidBandValue(value, bandMetadata)) {
           cellData[band] = tile.data.cells.numericProps[band].value[i];
           cellDataExists = true;
         }
@@ -108,4 +109,25 @@ function cellToChildrenSorted(parent: bigint, resolution: bigint): bigint[] {
       return tileA.x > tileB.x ? 1 : -1;
     }
   );
+}
+
+/**
+ * Returns true if the given value is valid (not NaN, not 'nodata')
+ * for the given raster band.
+ */
+function isValidBandValue(
+  value: unknown,
+  bandMetadata: RasterMetadataBand
+): value is number {
+  if (Number.isNaN(value)) {
+    return false;
+  }
+
+  // TODO(cleanup): Remove after API is updated to return 'nodata' as a number.
+  const nodata = bandMetadata.nodata;
+  if (typeof nodata === 'string') {
+    return Number(nodata) !== value;
+  }
+
+  return nodata !== value;
 }
--- a/src/filters/tileFeaturesRaster.ts
+++ b/src/filters/tileFeaturesRaster.ts
@@ -1,0 +1,111 @@
+import {
+  cellToChildren as _cellToChildren,
+  cellToTile,
+  geometryToCells,
+  getResolution,
+} from 'quadbin';
+import {RasterTile, SpatialFilter, Tile} from '../types.js';
+import {FeatureData} from '../types-internal.js';
+import {
+  RasterMetadata,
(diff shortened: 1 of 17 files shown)


src/widget-sources/widget-tileset-source.ts
```diff
@@ -106,9 +108,6 @@ export class WidgetTilesetSource extends WidgetSource<WidgetTilesetSourceProps>
     options,
   }: {
     spatialFilter: SpatialFilter;
-    // TODO(cleanup): As an optional property, 'uniqueIdProperty' will be easy to forget.
-    // Would it be better to configure it on the source function, rather than separately
-    // on the layer and in 'loadTiles()'?
```
donmccurdy: Removing this comment because `uniqueIdProperty` is inferred automatically in nearly all cases, the user shouldn't need to worry about it. In deck.gl this is a layer property, not a source/data property, and migrating that seems like too much of a breaking change for the benefit.

src/filters/tileFeaturesRaster.ts
```diff
@@ -0,0 +1,111 @@
...
+    for (let i = 0; i < children.length; i++) {
+      if (!spatialFilterCells.has(children[i])) continue;
+
+      const cellData: FeatureData = {};
+      let cellDataExists = false;
+
+      for (const band in tile.data.cells.numericProps) {
+        const value = tile.data.cells.numericProps[band].value[i];
+        // TODO(cleanup): nodata should be a number, not a string.
+        if (Number(bandMetadataByName[band].nodata) !== value) {
```
zbigg: no value can be NaN and this code will emits nans as proper data points

another one, builder interpret pixel as true non-rendered data if all bands used to calculate it are no data - it's tricky exception for rgb

donmccurdy: > no value can be NaN and this code will emits nans as proper data points

Hm I see – Probably NaN should be skipped here, even if it is not the 'nodata' value? I've made the changes in https://github.com/CartoDB/carto-api-client/pull/87/commits/ac50dfc84c140343ed4a88dab17eb4ec74954744.

> ...builder interpret pixel as true non-rendered data if all bands used to calculate it are no data

Here we are preparing the data for widget calculations, so the RGB display would not be affected, but ... I agree the same special cases will matter. Like if we are filtering or summing a band, some values could be missing.

I've tried to handle that by making raster widgets do exactly what a tileset widget would do, if pixels were replaced with quadbins – so there is no special case for rasters in the local widget calculations. If we want to do something differently than with tilesets, I think we'd need to discuss more the raster vs. tileset differences.

src/filters/tileFeaturesRaster.ts
```diff
@@ -0,0 +1,133 @@
...
+      if (!spatialFilterCells.has(children[i])) continue;
+
+      const cellData: FeatureData = {};
+      let cellDataExists = false;
+
+      for (const band in tile.data.cells.numericProps) {
+        const value = tile.data.cells.numericProps[band].value[i];
+        const bandMetadata = bandMetadataByName[band];
+
+        if (isValidBandValue(value, bandMetadata)) {
```
zbigg: Minor, performance-wise.

`nodata` can be string-encoded numeric value like `"-32768"`.
We should really preparse it to _number_ type and then use comparisons (`isValidBandValue` parses `nodada` to number internally if it happens to be string)
(We have hundreds of thousands of values here, no need to parse same number several hundred thousand times)

donmccurdy: Good call – done!

donmccurdy:
Merging to `feat/v0.5` branch.

PR #102, opened 2025-02-14
chore(lint): Set up ESLint

### LINKED ISSUE #101: Set up ESLint

donmccurdy:
Adds ESLint v9 and configures a linting check for production and non-production code in the repository. Some of the linter rules are looser than would be ideal (allowing implicit 'any',  for example) in order to keep the size of the initial PR a bit more manageable.

- fixes #101


eslint.config.js
```diff
@@ -0,0 +1,75 @@
...
+  '@typescript-eslint/no-explicit-any': 'off',
+  '@typescript-eslint/no-unsafe-argument': 'off',
+  '@typescript-eslint/no-unsafe-assignment': 'off',
+  '@typescript-eslint/no-unsafe-call': 'off',
+  '@typescript-eslint/no-unsafe-member-access': 'off',
+  '@typescript-eslint/no-unsafe-return': 'off',
+
+  // TODO: These rules should be re-enabled for all code.
+  '@typescript-eslint/ban-ts-comment': 'off',
+};
```
jmgaya, line 34: Thinking out loudly, I think we should start thinking about creating some kind of `@carto/eslint-config`, because I think codebases should be aligned within these rules, and maintaining only an ESLint Config seems easier, safer, and better, for all of us, WDYT?

Said so, thank you very much for caring about DEV/PROD rules, I think this aligns pretty well with the mess in `cloud-native` where we disable the same ESLint rules on every test, great job! ❤️

jmgaya (approved):
Great job 💪 !

eslint.config.js
donmccurdy: Perhaps! I don't feel too strongly about a CARTO-specific config. I think in this repo the default ESLint + TypeScript-ESLint rules are basically OK, but in App Templates we did need to customize with new plugins for the various frameworks:

- https://github.com/CartoDB/carto-app-templates/pull/70/files

If we have the cloud-native monorepo updated to ESLint v9 with a shared common config file (could just be a local file in the repo) then at least we're 90% of the way there and could decide if that config makes sense to use in other projects outside the monorepo.

PR #106, opened 2025-02-19
fix(build): Enable compatibility with moduleResolution = nodenext

### LINKED ISSUE #99: Imports should use explicit .js extensions for moduleResolution=nodenext compatibility
Imports in published `.d.ts` files must point to `.js` extensions, and TypeScript has for a very long time been against handling that renaming itself:

- https://github.com/microsoft/TypeScript/issues/49462

The filenames are supposed to be paths that can be resolved _after_ TypeScript compilation, so we are stuck with either `./Filter` or `./Filter.js`. Currently the codebase has a mix of both ... so that is not great, I would like to clean it up to be consistent one way or the other. 😅

While I'm happy with either in an _application_, I think that for _libraries_ we may need to use `.js` — because that's what should appear in the published `.d.ts` files. It's required for customers using certain configurations, particularly `moduleResolution: 'nodenext'`. I've had this feedback in some of my personal repositories in the past, longer discussion here:

- https://github.com/donmccurdy/glTF-Transform/pull/827

donmccurdy:
- Fixes #99

Customers using TypeScript configured to `moduleResolution="nodenext"` or `moduleResolution="node2016"` will run into the problems described at:

- https://github.com/donmccurdy/glTF-Transform/pull/827

With our current build system (and most others today) type definitions in the `build/` outputs contain one `.d.ts` file per source file, and these files *must* refer to one another by `.js` extensions when consumed by any TypeScript project using `moduleResolution="nodenext"`. This could be avoided if we had a build system that bundled all type definitions into a single `.d.ts` file — in which case there are no imports in the generated types — but I'm not confident recommending a build system with this feature.

TypeScript does not rename extensions from `./foo` or `./foo.ts` to `./foo.js` by itself (https://github.com/microsoft/TypeScript/issues/49462). 

deck.gl has [a special build step](https://github.com/uber-web/ocular/blob/6d84ec50d8c80dc2cf361e818e9edb35f597042a/modules/dev-tools/src/ts-plugins/ts-transform-append-extension/index.ts) configured to do this renaming.

I'd rather not complicate the build process, so I think the simplest and safest solution is to ensure that all relative imports are `.js` in the source files, so they'll already be correct when type definitions are built. By adopting `moduleResolution="nodenext"` for this library we can be sure that the extensions will be consistent, because `yarn lint` enforces it. For examples and tests it doesn't particularly matter either way.


felixpalmer (approved):
My preference would be to copy what deck does, at it is quite messy to have those '.js' extensions everywhere and I wonder if IDEs will get confused. But OK to go with this solution for now

donmccurdy:
As weird as it looks, this does seem to be what [TypeScript recommends](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext):

>  Because node16 and nodenext are the only module options that reflect the complexities of Node.js’s dual module system, they are the only correct module options for all apps and libraries that are intended to run in Node.js v12 or later, whether they use ES modules or not.

But ... I do not like TypeScript's choices around bundling at all, so pretty open to changing this. The only other way I've handled this before is [rollup-plugin-dts](https://github.com/Swatinem/rollup-plugin-dts), which worked fine but would mean a second build step. Have heard good things about [tsup](https://github.com/egoist/tsup), which appears to handle this out of the box, but I haven't used tsup yet personally.

PR #72, opened 2025-01-15
feat(rasters): Support widget calculations on rasters

donmccurdy:

PR #108, opened 2025-02-20
feat(tileset): Use Web Workers

donmccurdy:
Adds support for Web Workers in tileset sources and (when #72 is merged) rasters.

Tested in:

- [x] Vite
- [x] Webpack 5
- [ ] ❌ Parcel 2

Parcel fails at the build step with two different errors, it doesn't like our using string literal workers to _test_ for Web Worker support...

CleanShot 2025-02-25 at 17 44 03@2x

... and also doesn't like our using the package name to create the real Web Worker:

CleanShot 2025-02-25 at 17 44 35@2x

Parcel does have [documentation about Web Worker support](https://parceljs.org/languages/javascript/#web-workers), though, so I'm going to keep looking for a fix on that.

One other new addition is a `.destroy()` method on WidgetSource instances. Web Workers are expensive to create, and browsers [limit how many can be active](https://stackoverflow.com/questions/13574158/number-of-web-workers-limit), so we do need to clean those up. It's a bit awkward to have the WidgetSource created implicitly by vectorTilesetSource, but then `.destroy()` _must_ be called explicitly...

```javascript
  if (data?.widgetSource) {
    await data.widgetSource.destroy();
  }

  data = await vectorTilesetSource({
    tableName: 'my_table',
    accessToken: '•••',
    connectionName: '...',
  });
```

... so we can discuss if there's a better way to handle this.

***

Old notes —

Open questions:

- [x] ~Avoid starting Web Workers until a widget request is made. We don't want to spawn workers if the user is just rendering the tileset.~
- [x] ~How many Web Workers should be spawned? A singleton? One per dataset? A pool of N workers, with datasets balanced among them?~ One per `vectorTilesetSource` call. Rebuilding the source will rebuild the worker (if  widget APIs are being used).
- [x] ~If we're using ES Module workers, is that compatible with CJS builds?~ Workers will be ESM-only.
- [x] ~If workers are opt-in or opt-out, what's the API to switch?~ Opt-out, by `widgetSourceWorker: false` option.
- [x] ~If Microbundle generates named chunks like `widget-tileset-worker-BXxfh1Ze.js`, is that compatible with `package.json#exports`?~ Replaced microbundle with tsup.

These issues could probably be solved like deck.gl does, by publishing worker code to a CDN and loading that... But this makes builds and development more complicated, and CDNs may fail, I'd prefer to avoid the CDN.

The last version of Node.js that doesn't support [`require(esm)`](https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js/) will reach end-of-life in April... maybe we can say Web Workers are available only for ESM builds? We can use an extra entrypoint (`@carto/api-client/worker`) for the worker code.


src/widget-sources/widget-tileset-worker-source.ts
```diff
@@ -0,0 +1,238 @@
...
+    // TODO: ViewState may contain non-serializable data, which we do not need.
+    // Remove this sanitization after sc-469614 is fixed.
+    const options = params[0] as any;
+    if (options?.spatialIndexReferenceViewState) {
+      const {zoom, latitude, longitude} =
+        options.spatialIndexReferenceViewState;
+      options.spatialIndexReferenceViewState = {zoom, latitude, longitude};
+    }
+
+    worker.postMessage({
```
zbigg: I'm not sure about ordering guarantees, but sending message and _then_ only installing listener seems awkward.

I think it will be safer and more intuitive to instal listener first, then post message

src/widget-sources/widget-tileset-worker-source.ts
```diff
@@ -0,0 +1,238 @@
...
+    } as WorkerRequest);
+
+    return new Promise((resolve, reject) => {
+      function listener(e: MessageEvent) {
+        const response = e.data as WorkerResponse;
+        if (response.requestId !== requestId) return;
+
+        worker.removeEventListener('message', listener);
+
+        if (signal?.aborted) {
```
zbigg: to be robust, we should also install `signal.onabort(cleanup)` to abort and unsubscribe immediately when we abort ... and not to wait for response and only then abort (not sure how long it will take)

src/widget-sources/widget-tileset-worker-source.ts
```diff
@@ -0,0 +1,238 @@
...
+    } as WorkerRequest);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/require-await
+  override async getFeatures(): Promise<FeaturesResponse> {
+    throw new Error('getFeatures not supported for tilesets');
+  }
+
+  async getFormula({
+    abortController,
```
zbigg: I know it's late, but it's good practice that 'consuming' APIs are receiving only signal (like fetch, like [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal))
Only owners should have references to AbortController. This makes seperation of concern more visible.

I see no reason for _request_ to control external signal by itself.

zbigg:
> We can use an extra entrypoint (@carto/api-client/worker)

That is approach we used in old library and i believe it's best _started_ at least.

There are several approaches to embed worker into library, most of them have problems. This one seems easiest from our PoV, but adds some burden to users.

1) embed worker code as string - requires `unsafe-eval` CSP (afair)
2) load blob from CDN - again, cross origin problems but better than eval (deck does this?)
3) force user to configure his app to create worker bundle explicitly and pass it in some initializer
4) If creating UMD-like bundles for usage as `<script>` tag from CDN, you can detect "main" bundle base-URL and find sibling worker bundle URL ... but that's not option for us ...

zbigg:
For 3 to work, we should either allow users to determine worker script URL by themselves or let tool do it.

First would be option: "workerUrl" which is given enables workers and users are asked to bundle `@carto/api-client/worker` by themselves.

Second (vite compatible?) would be to ask user for option "worker: Worker", where they would just add 
```
const cartoApiClientWorker = new Worker(new URL('@carto/api-client/worker', import.meta.url))
``` 
and pass it to our API.

Does it even work to create worker bundle directly from "external living in node_modules"?

donmccurdy:
> This one seems easiest from our PoV, but adds some burden to users...

@zbigg I don't think I understand what burden is added – the approach in this PR should "just work" (no `unsafe-eval`, no CDN, no configuration) in modern bundlers — Vite, Parcel 2+, Webpack 5+ — _if_ the application is using ESM. But, in older bundlers or with CommonJS, the widgets will need to be calculated without Web Workers, which I'm hoping is OK.

So the approach here is not really any of 1, 2, 3, or 4, I think. Do you prefer one of those?

src/widget-sources/widget-tileset-worker-source.ts
donmccurdy: Good point! I've reordered it.

src/widget-sources/widget-tileset-worker-source.ts
donmccurdy: Ok, I like that idea thanks! Added.

src/widget-sources/widget-tileset-worker-source.ts
donmccurdy: Hm, yes that's a good point. I'd copied this from C4R but I guess it was more of an "internal" detail then, but here it is a public API for external users and should be clean and consistent like [using an AbortController with the fetch() API](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#aborting_a_fetch_operation_using_an_explicit_signal). 

I think this could be a good time to make the change, I'd like to bump the package from a v0.4 to v0.5 release anyway, since adding the Web Workers might be considered a breaking change. Let's double-check with @srtena but I think we can switch from this...

```javascript
const abortController = new AbortController();

const result = await widgetSource.getFormula({ ..., abortController });
```

... to this ...

```javascript
const abortController = new AbortController();
const signal = abortController.signal;

const result = await widgetSource.getFormula({ ..., signal });
```

... which would be the same as `fetch()`.

donmccurdy: Fixed in a new PR:

- https://github.com/CartoDB/carto-api-client/pull/110

examples/03-tileset/app.ts
```diff
@@ -59,6 +59,10 @@ await updateSources();
  */
 
 async function updateSources() {
+  if (data?.widgetSource) {
+    await data.widgetSource.destroy();
```
padawannn: Have you considered using `FinalizationRegistry` in the `widgetSource` code to don't need to call this destroy manually? Something like:

```
const workerRegistry =
  typeof FinalizationRegistry !== 'undefined'
    ? new FinalizationRegistry((worker: Worker) => {
        console.warn('Object collected without calling destroy(). Terminate worker.');
        worker.terminate();
      })
    : null;

export class WidgetTilesetWorkerSource extends WidgetSource<WidgetTilesetSourceProps> {
  protected _worker: Worker | null = null;

  constructor(props: WidgetTilesetSourceProps) {
     ...
    workerRegistry?.register(this, this._worker);
  }
```

Of course, we could keep the manual destroy call possibility

padawannn (changes requested):

examples/03-tileset/app.ts
donmccurdy: I've done some testing with FinalizationRegistry — this would be nice as a safeguard if the user forgets to destroy the source! — but for some reason I could never get the cleanup callback to run. I'm not sure if that's because something kept the WidgetSource from being garbage collected after deleting all references to it, or if GC just doesn't run at all if there isn't much memory being allocated by the application.

So I think I agree – we'll need the manual .destroy(), and mayyybe also can add the FinalizationRegistry as a backup.

donmccurdy:
I've refactored this PR a fair bit. The main change is to NOT to have two separate `WidgetTilesetSource` vs. `WidgetTilesetWorkerSource` classes exposed to users. 

Now, instead, there's the public `WidgetTilesetSource` class, and an (internal) `WidgetTilesetSourceImpl` class with the in-memory calculation logic. Both are implementations of `WidgetSource`, but `WidgetTilesetSource` defers all calculations to a `WidgetTilesetSourceImpl` instance — which could be either on the main thread or in a worker. 

Example:

```js
// local
const { widgetSource } = vectorTilesetSource({..., widgetSourceWorker: false});
const widgetSource = new WidgetTilesetSource({..., widgetSourceWorker: false});

// worker
const { widgetSource } = vectorTilesetSource({..., widgetSourceWorker: true});
const widgetSource = new WidgetTilesetSource({..., widgetSourceWorker: true});
```

I tried a number of things to extend support for different combinations of environment/format/bundler, but I hope it's "good enough" for now. In case we need to improve or document support later, I've written up the compatibility (and possible future changes) in a new issue:

- https://github.com/CartoDB/carto-api-client/issues/112

padawannn:
> I've refactored this PR a fair bit. The main change is to NOT to have two separate `WidgetTilesetSource` vs. `WidgetTilesetWorkerSource` classes exposed to users.
> 
> Now, instead, there's the public `WidgetTilesetSource` class, and an (internal) `WidgetTilesetSourceImpl` class with the in-memory calculation logic. Both are implementations of `WidgetSource`, but `WidgetTilesetSource` defers all calculations to a `WidgetTilesetSourceImpl` instance — which could be either on the main thread or in a worker.
> 
> Example:
> 
> ```js
> // local
> const { widgetSource } = vectorTilesetSource({..., widgetSourceWorker: false});
> const widgetSource = new WidgetTilesetSource({..., widgetSourceWorker: false});
> 
> // worker
> const { widgetSource } = vectorTilesetSource({..., widgetSourceWorker: true});
> const widgetSource = new WidgetTilesetSource({..., widgetSourceWorker: true});
> ```
> 
> I tried a number of things to extend support for different combinations of environment/format/bundler, but I hope it's "good enough" for now. In case we need to improve or document support later, I've written up the compatibility (and possible future changes) in a new issue:
> 
> * [Web Worker support #112](https://github.com/CartoDB/carto-api-client/issues/112)

Looks well @donmccurdy . On the other hand what about to add the `FinalizationRegistry` as a backup as you said in your previous comment?

.github/workflows/release.yml
```diff
@@ -68,4 +68,4 @@ jobs:
       - name: Publish
         env:
           DIST_TAG: ${{ steps.dist-tag.outputs.result }}
-        run: yarn npm publish --tag ${DIST_TAG}
+        run: yarn npm publish --tag ${DIST_TAG} --access public
```
jmgaya: I think you've already mentioned it somewhere, but is @vmilan aware of this? If I'm correct the idea was to share the release process between both repositories, so I guess we would like to have this there too?

src/widget-sources/widget-tileset-source-impl.ts
```diff
@@ -0,0 +1,417 @@
+/* eslint-disable @typescript-eslint/require-await */
```
jmgaya, line 1: There are some scenarios that may be worth being covered with Tests, @donmccurdy do we have plans to do so? e.g. `getFormula` or `getTable`

src/widget-sources/widget-tileset-source-impl.ts
```diff
@@ -0,0 +1,417 @@
...
+      min: aggregationFunctions.min(filteredFeatures, column),
+      max: aggregationFunctions.max(filteredFeatures, column),
+    };
+  }
+
+  /****************************************************************************
+   * INTERNAL
+   */
+
+  private _getFilteredFeatures(
```
jmgaya, line 374: This is completely unrelated to this PR, but have we thought about using `#` if we want to make sure our consumers do not access this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

If we just want this at TypeScript level, then I guess we're fine 👍

jmgaya (approved):
Amazing work!

.github/workflows/release.yml
donmccurdy: Meridian has the configuration here, either is equivalent:

https://github.com/CartoDB/meridian-ds/blob/37a2d7de6e1448f4c7113c46db75abd66010b96d/package.json#L128-L130

I'd originally included that in `@carto/api-client` but later removed it in https://github.com/CartoDB/carto-api-client/pull/109, not realizing that `yarn npm publish` had different defaults from `npm publish` and would need this. Had to restore it here because otherwise I can't make alpha releases to test things, but let's just revert #109, I've opened a PR to revert back so we don't need to update any other projects:

- https://github.com/CartoDB/carto-api-client/pull/114

***

Aside: I'm a little worried about having 'shared' configuration and documentation in each repository — for app templates, quadbin-js, carto-api-client, and meridian — because for any small change in 1 project, someone has to update the other 3, and wait for code reviews. 😓  

So if we really want to share same pattern then I think we need to invest time setting up a shared Github Action and a shared documentation location. Totally fine if we don't want to do that yet — abstraction can sometimes be harder to maintain than a little duplication — but then we should not worry too much about keeping everything in sync manually.

src/widget-sources/widget-tileset-source-impl.ts
donmccurdy: I couldn't convince GitHub to show the diff as a rename, but this file is really just renamed `widget-tileset-source.ts`, so these functions do have coverage! Currently `test/widget-sources/widget-tileset-source.test.ts` provides the test coverage for both WidgetTilesetSource and WidgetTilesetSourceImpl, since the 'impl' is used internally, and we can't run the Workers in Node.js right now.

In any case — there were a few other widget functions missing coverage, so I've added most of those now (all but getTimeSeries).

Also fixed `yarn coverage` which had broken with the build tool replacement, so good callout on coverage!

src/widget-sources/widget-tileset-source-impl.ts
donmccurdy: I think my slight preference is just "TypeScript private" — it is nice to debug more easily at runtime, I don't personally care if consumers access this, as long as they realize the risk, which the `_` prefix hopefully communicates.

donmccurdy:
> ...On the other hand what about to add the FinalizationRegistry as a backup as you said in your previous comment?

@padawannn I still like this idea but I haven't gotten it to work in any tests yet. So to avoid  adding what might be dead code — maybe Web Worker event handlers prevent garbage collection? — I think let's leave this for a separate PR if/when we can figure out at least one sequence of events where the GC callback runs. 

I think your earlier review is still blocking a merge here, so just let me know if there are blocking issues left, thanks!

src/widget-sources/widget-tileset-source-impl.ts
donmccurdy: Current coverage in `src/widget-sources/*`, the last one is harder with workers:

File                                    | % Stmts | % Branch | % Funcs | % Lines |
----------------------------------------|---------|----------|---------|---------|
widget-query-source.ts                |     100 |      100 |     100 |     100 |
widget-remote-source.ts               |   98.33 |    76.66 |     100 |    98.3 |
widget-source.ts                      |      70 |    33.33 |   66.66 |   66.66 |
widget-table-source.ts                |     100 |      100 |     100 |     100 |
widget-tileset-source-impl.ts         |   72.15 |    60.34 |   78.94 |   71.79 |
widget-tileset-source.ts              |   28.57 |    29.16 |      50 |   29.41 |

.github/workflows/release.yml
jmgaya: Absolutely! I think we don't have enough "hands to cook this cake" 😄 

When I said "share", I was meaning to "have something similar", not to have some kind of "common place" where all these processes are documented and/or hosted 👍

src/widget-sources/widget-tileset-source-impl.ts
jmgaya: Sure! Just wanted to highlight that, thanks for your explanation 😄

donmccurdy:
Merging to `feat/v0.5` branch.

PR #109, opened 2025-02-24
chore(npm): Remove package.json#packageConfig

donmccurdy:
This config was needed during initial development, but now that the package is published with stable releases, it can be removed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,10 +10,6 @@
   "packageManager": "yarn@4.3.1",
   "version": "0.5.0-alpha.2",
   "license": "MIT",
-  "publishConfig": {
-    "access": "public",
-    "tag": "alpha"
-  },
   "type": "module",
   "sideEffects": false,
   "source": "src/index.ts",


padawannn (approved):

donmccurdy:
Found a problem with this change: `npm publish` needs `--access public` [only the _first_  time](https://docs.npmjs.com/cli/v9/using-npm/scope#publishing-public-scoped-packages-to-the-primary-npm-registry) that a scoped package is published, but with `yarn npm publish` we must specify `--access public` (or equivalent `package.json#packageConfig`) [_every_ time](https://yarnpkg.com/cli/npm/publish) we publish the package. Since we are using `yarn npm publish` we'll need to either restore this config, or add `--access public` in release.yml here:

https://github.com/CartoDB/carto-api-client/blob/3e57d1b85292403e49ab4f1950266a5198438fd2/.github/workflows/release.yml#L71

I'll open a PR.

donmccurdy:
Planning a revert in #114.

PR #110, opened 2025-02-24
refactor(widgets): Replace option 'abortController' with 'signal'

donmccurdy:
For consistency with [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#aborting_a_fetch_operation_using_an_explicit_signal), replaces the 'abortController' option with 'signal'. Example:

```javascript
const abortController = new AbortController();
const signal = abortController.signal;
const result = await widgetSource.getFormula({ ..., signal });
```

I'd like to release support for tilesets, rasters, and Web Workers as part of a v0.5 release, so this is a safer time for renaming, since npm clients will not auto-upgrade from `^0.4` to `v0.5`.

Context:

- https://github.com/CartoDB/carto-api-client/pull/108#discussion_r1967536332

### DIFF
--- a/src/models/common.ts
+++ b/src/models/common.ts
@@ -4,7 +4,7 @@ import {InvalidColumnError} from '../utils.js';
 export interface ModelRequestOptions {
   method: 'GET' | 'POST';
   headers?: Record<string, string>;
-  abortController?: AbortController;
+  signal?: AbortSignal;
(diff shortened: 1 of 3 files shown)


padawannn (approved):

jmgaya (approved):
Great upgrade! This library is just looking awesome ❤️

PR #122, opened 2025-03-05
fix(query,sources): Ensure consistent setClient() support

donmccurdy:
Changes:

- Fix `setClient()` support in `query()`
- Fix `setClient()` support in TileJSON requests (was not propagated from user-specified params)
- Improve test coverage

### DIFF
--- a/test/setup.ts
+++ b/test/setup.ts
@@ -11,9 +11,6 @@ const DEFAULT_CLIENT_ID = getClient();
 
 afterEach(() => {
   setClient(DEFAULT_CLIENT_ID);
-});
-
-afterEach(() => {
   vi.restoreAllMocks();
   vi.unstubAllGlobals();
 });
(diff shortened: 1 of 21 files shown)


test/setup.ts
```diff
@@ -0,0 +1,14 @@
...
+// need to disable parallelism or isolation for debugging purposes. To confirm
+// that all tests pass without parallism or isolation, use:
+//
+// yarn test --no-file-parallelism --no-isolate
+
+const DEFAULT_CLIENT_ID = getClient();
+
+afterEach(() => {
+  setClient(DEFAULT_CLIENT_ID);
+});
```
jmgaya, line 16: Now that we've created a shared file for Tests, maybe we can also reset mocks, to prevent false positives 🤔 ?

jmgaya (approved):
Great job ❤️

test/setup.ts
donmccurdy: Good point — done!

Issue #124, opened 2025-03-11
fetchMap tracker


felixpalmer:
Tracker for PRs needed to move fetchMap to carto-api-client

### Port existing code from `deck.gl/carto`:

- [x] Copy across existing test app from deck.gl repo https://github.com/CartoDB/carto-api-client/pull/125
- [x] Copy fetchMap implementation from deck.gl repo https://github.com/CartoDB/carto-api-client/pull/126
- [x] Port tests https://github.com/CartoDB/carto-api-client/pull/127
- [x] Remove document mode layers & add LayerProvider to decouple from deck.gl https://github.com/CartoDB/carto-api-client/pull/130
- [x] Remove deck.gl dependency https://github.com/CartoDB/carto-api-client/pull/136
- [x] Merge all PRs from above to main
- [x] Expose popup settings https://github.com/CartoDB/carto-api-client/pull/133
- [x] Default to `PointLabelLayer` in `VectorTileLayer` https://github.com/visgl/deck.gl/pull/9524 & https://github.com/CartoDB/carto-api-client/pull/137

### Changes from Builder

- [x] getCartoSource https://github.com/CartoDB/carto-api-client/pull/138
- [x] Filtering (server & client) https://github.com/CartoDB/carto-api-client/pull/139

### Legends

- [x] Legend example https://github.com/CartoDB/carto-api-client/pull/154
- [x] Expose `scales` for building legend https://github.com/CartoDB/carto-api-client/pull/179

### Cleanup

- [x] Remove `GeojsonResult | JsonResult` https://github.com/CartoDB/carto-api-client/pull/138#discussion_r2003422354
- [x] Tidy `getSpatialIndexFromGeoColumn` https://github.com/CartoDB/carto-api-client/pull/138/files/ea16798786ac41320537930c421c42b8d7cd928d#r2003465870 https://github.com/CartoDB/carto-api-client/pull/155
- [x] Pull in more complete types from Builder https://github.com/CartoDB/carto-api-client/pull/138#discussion_r2003673968
- [x] Remove `fetchMap` from deck.gl

AnaDiaz85:
List of broken maps:
- 'b8abc46c-3c7f-489f-b16f-0664872ad82a', // Vector - Table - Snowflake - 74k bike accidents France

// Test QA

[...]

PR #125, opened 2025-03-11
Copy fetchMap test app from deck.gl repo

felixpalmer:

#### Background

To avoid huge diffs, as a first step I'm copying across the test app with minimal changes

#### Change list

- Add new test app (direct copy)
- Use Maplibre rather than Mapbox
- Remove `createMapWithDeckController()`
- Lint fixes


package.json
```diff
@@ -109,6 +114,9 @@
     "vite": "^6.1.0",
     "vitest": "3.0.7"
   },
+  "peerDependencies": {
+    "@deck.gl/carto": "~9.1.0"
+  },
```
donmccurdy: I don't think we need this as a peer dependency – or at least not for this PR. And adding it will bring in dependencies on much of deck.gl. Could we wait to add this until we're sure we need it, so we can still publish standalone releases in the meantime?

examples/05-fetchMap/app.ts
donmccurdy, line 1: I assume we'll probably want to keep both copies of this example for debugging? One in deck.gl and one here?

donmccurdy (approved):

examples/05-fetchMap/app.ts
donmccurdy, line 1: Could we name the directory `05-fetch-map` for consistency with the other kebab-case folders?

package.json
felixpalmer: Yep, you're right - I was using Javi's test app as a starting point and copied this over by accident

examples/05-fetchMap/app.ts
felixpalmer: Yes, I think so. It is very useful to be able to pull a Builder map into a deck.gl dev environment

PR #126, opened 2025-03-12
Move current fetchMap implementation from deck.gl repo

felixpalmer:

#### Change list

- Copy across files from deck.gl repository
- Update imports to use local paths & style
- Simplest possible typescript fixes to resolve stricter rules in this repo
- Update test app to use local implementation


src/fetchMap/basemap-styles.ts
donmccurdy, line 1: Could we name the directory `src/fetch-map` for consistency with others?

src/fetchMap/utils.ts
```diff
@@ -0,0 +1,61 @@
+import {log} from '@deck.gl/core';
```
donmccurdy: Possibly unused? Currently I think the ESLint CI isn't running because this PR doesn't target the main branch, maybe that's something we should change.

donmccurdy:
Wondering about our migration path for dependencies here, if we do...

1. copy fetchMap as-is
2. publish releases
3. update fetchMap to not depend on `@deck.gl/*` (except perhaps types)
4. publish releases

... then we have some intermediate releases with heavy extra dependencies and possibly some compatibility challenges for Node.js-based apps. Is that OK? Or should we be keeping this on a feature branch until we reach step (3)? Perhaps keep a release branch for maintenance updates between steps (1) and (3) ?

src/fetchMap/utils.ts
felixpalmer: Ah, I missed that. If you could update the CI to run on all branches that would be useful

felixpalmer:
> Or should we be keeping this on a feature branch until we reach step (3)?

I certainly think we should avoid any releases until we reach step (3) I don't think this should take too long, I hope to tackle this tomorrow and we don't need to merge this PR until the second part is ready

donmccurdy:
Ok thanks – if we think we'll have (3) in the next few days then let's merge it all into `main` at the same time, I'll likely need to make a release in the meantime for #128.

donmccurdy (approved):
Continuing from the Slack discussion I think it'd be great if we could drop the `@deck.gl/*` dependencies — at least initially — so that there's no risk this is a breaking change, returning `unknown` or `any` if necessary. Not sure which PR change will land in, and don't want to block the whole stack, so just approving this ahead of time.

Thanks @felixpalmer!

PR #127, opened 2025-03-12
Port over fetchMap tests

felixpalmer:

#### Change list

- Copy across tests from deck.gl
- Port tests to vitest
- Remove parseMap tests that we too much hassle to port and will likely change anyway due to the deprecation of document mode


src/fetchMap/index.ts
```diff
@@ -1,4 +1,7 @@
 export {default as BASEMAP} from './basemap-styles.js';
 export {fetchMap} from './fetch-map.js';
 export type {FetchMapOptions, FetchMapResult} from './fetch-map.js';
-export type {Basemap, MapLibreBasemap, GoogleBasemap} from './types.js';
+export type {Basemap, MapLibreBasemap, GoogleBasemap, KeplerMapConfig} from './types.js';
+
+export * from './layer-map.js';
+export * from './parse-map.js'
```
donmccurdy: Optional - if something is included only for testing I'm trying to add `@internal` TSDoc annotations.

test/fetchMap/basemap.spec.ts
donmccurdy, line 1: Nit: Perhaps `test/fetch-map`?

donmccurdy (approved):

src/fetchMap/index.ts
felixpalmer: Do you know if there is a way to avoid having to export these at all? In the deck.gl repo it is possible to "reach into" the source, like this: https://github.com/visgl/deck.gl/blob/master/test/modules/carto/api/basemap.spec.ts#L8

donmccurdy: I believe we could configure Vitest and tsconfig.json to allow that, too. The downside is that we'd be unit testing uncompiled code, and very occasionally that could matter.

Perhaps the alternative would be to do something like https://github.com/antfu/vitest-package-exports, and test that any changes to internal vs. non-external package exports are 'expected'?

PR #128, opened 2025-03-12
feat(widgets): Remove spatialIndexReferenceViewState param

donmccurdy:
Removes the `spatialIndexReferenceViewState` parameter, previously required for backend widget calculations with H3 and Quadbin spatial index sources. The parameter was used to compute spatial filter resolution in the selected spatial index scheme, but now resolution is determined automatically based on the size of the spatial filter — not the viewport.

For backward compatibility, type definitions are kept but deprecated. Last reference will be removed in v0.5.

I'll make an alpha release and test this in Builder before merging and publishing a stable v0.4.9 release.

### DIFF
--- a/src/widget-sources/types.ts
+++ b/src/widget-sources/types.ts
@@ -24,8 +24,6 @@ interface BaseRequestOptions {
   signal?: AbortSignal;
   spatialFilter?: SpatialFilter;
   spatialFiltersMode?: SpatialFilterPolyfillMode;
-  /** Required for table- and query-based spatial index sources (H3, Quadbin). */
-  spatialIndexReferenceViewState?: ViewState;
   /** Overrides source filters, if any. */
   filters?: Filters;
   filterOwner?: string;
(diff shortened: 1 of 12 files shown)


src/widget-sources/types.ts
```diff
@@ -24,7 +24,12 @@ interface BaseRequestOptions {
...
   spatialFilter?: SpatialFilter;
   spatialFiltersMode?: SpatialFilterPolyfillMode;
-  /** Required for table- and query-based spatial index sources (H3, Quadbin). */
+  /**
+   * Deprecated parameter previously used for H3 and Quadbin widgets. Now has
+   * no effect and will be removed in a future version.
+   * @deprecated Parameter has no effect.
+   * @todo TODO(v0.5): Remove spatialIndexReferenceViewState parameter.
+   */
   spatialIndexReferenceViewState?: ViewState;
```
donmccurdy, line 33: Keeping this type definition for backcompat until v0.5.

zbigg (approved):

PR #130, opened 2025-03-13
fetchMap: remove document mode and deck.gl dependencies

felixpalmer:

#### Change list

- Replace `TILE_LAYER_TYPE_TO_LAYER` with injected `LayerProvider`
- Remove document mode layers
- Remove any @deck.gl imports (except for types)
- Remove `PointLabelLayer` & `CollisionFilterExtension`


src/fetch-map/layer-map.ts
```diff
@@ -13,24 +13,19 @@ import {
...
+
+export type LayerProvider = {
+  clusterTile: ConstructorOf<ClusterTileLayer>,
+  h3: ConstructorOf<H3TileLayer>,
+  heatmapTile: ConstructorOf<HeatmapTileLayer>,
+  mvt: ConstructorOf<VectorTileLayer>,
+  quadbin: ConstructorOf<QuadbinTileLayer>,
+  raster: ConstructorOf<RasterTileLayer>,
+  tileset: ConstructorOf<VectorTileLayer>
+}
```
donmccurdy: Do the keys of this object come from somewhere particular? I might be inclined to make these particular types less strict...

```typescript
// (A)
export type LayerProvider = Record<'clusterTile' | 'h3' | ..., ConstructorOf<Layer>>;

// (B)
export type LayerProvider = Record<'clusterTile' | 'h3' | ..., unknown>;
```

... otherwise TS is potentially doing deep compatibility checks here that will just make version upgrades, layer overrides, and dual package hazard more risky, without much type safety benefit.

src/fetch-map/parse-map.ts
```diff
@@ -88,8 +78,10 @@ export function parseMap(json: any) {
             assert(data, `No data loaded for dataId: ${dataId}`);
             const {Layer, propMap, defaultProps} = getLayer(
               type as LayerType,
+              // @ts-ignore
```
donmccurdy, line 81: Optional - Would a cast be appropriate here? This looks like it 'should' be OK?

donmccurdy (approved):
Two comments, both optional -- thanks!

src/fetch-map/layer-map.ts
felixpalmer: The keys are based on what the API gives us, via the `keplerMapConfig` (with CARTO extensions). I've gone ahead and loosened the value to `Layer`. This gives us the option to swap out an arbitrary `Layer` and just have it receive the props, thus not forcing the consumer to subclass a CARTO layer if they don't want to (although I expect most will)

felixpalmer: The extra benefit is that we don't need to have `@deck.gl/carto` as a peerDependency and we can just use `@deck.gl/core` which is more lightweight

examples/05-fetch-map/app.ts
```diff
@@ -81,7 +94,7 @@ async function createMap(cartoMapId: string) {
   }
 
   // Get map info from CARTO and update deck
-  const result = await fetchMap(options);
+  const result = await fetchMap(options, layerProvider);
```
donmccurdy, line 97: I was imagining that `layerProvider` would be part of the options parameter – any particular reason to make it a new parameter?

felixpalmer: As discussed elsewhere, we will remove it altogether in a follow up PR

PR #135, opened 2025-03-18
fetchMap: tests & remove document mode and deck.gl dependencies

felixpalmer:
Github was not clever enough to update the base branch and so I accidentally merged https://github.com/CartoDB/carto-api-client/pull/127 & https://github.com/CartoDB/carto-api-client/pull/130 into `felix/fetchmap-implementation`


test/fetch-map/layer-map.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+        colorRange: {
+          colors,
+          colorMap: [
+            [0, '#E3611C'],
+            [1, '#F1920E'],
+            [5, '#FFC300'],
+          ],
+        },
+        opacity: 1,
+        data: [{v: 0}, {v: 1}, {v: 5}],
```
zbigg: those with `data: [...]` (anything not tilejson) are for document layers. arent't they? If yes, i think we can remove them.

zbigg (commented):
`calculateDomain` two last branches are for document mode only (?)

jmgaya (approved):

PR #138, opened 2025-03-19
Replace _fetchMapDataset with getCartoSource

felixpalmer:

#### Change list

- Copy in logic from [getCartoSource](ttps://github.com/CartoDB/cloud-native/blob/main/workspace-www/src/features/common/utils/cartoDeckGL.ts#L79)
- Tidy up types

### DIFF
--- a/src/constants-internal.ts
+++ b/src/constants-internal.ts
@@ -19,6 +19,9 @@ export const DEFAULT_MAX_LENGTH_URL = 7000;
 /** @privateRemarks Source: @deck.gl/carto */
 export const DEFAULT_TILE_RESOLUTION = 0.5;
 
+/** @internal */
+export const REDUCED_QUERIES_TILE_RESOLUTION = 1;
+
 /**
  * @privateRemarks Source: @deck.gl/carto
  * @internal
@@ -30,3 +33,7 @@ export const DEFAULT_AGGREGATION_RES_LEVEL_H3 = 4;
  * @internal
  */
 export const DEFAULT_AGGREGATION_RES_LEVEL_QUADBIN = 6;
+
+/** @internal */
+export const DEFAULT_AGGREGATION_EXP_ALIAS = '__aggregationValue';
+export const DEFAULT_AGGREGATION_EXP = `1 AS ${DEFAULT_AGGREGATION_EXP_ALIAS}`;
--- a/src/constants-internal.ts
+++ b/src/constants-internal.ts
@@ -19,7 +19,10 @@ export const DEFAULT_MAX_LENGTH_URL = 7000;
 /** @privateRemarks Source: @deck.gl/carto */
 export const DEFAULT_TILE_RESOLUTION = 0.5;
 
-/** @internal */
+/**
+ * @todo TODO(v0.5): Update DEFAULT_TILE_RESOLUTION and remove this constant.
+ * @internal
+ */
 export const REDUCED_QUERIES_TILE_RESOLUTION = 1;
 
 /**
--- a/src/fetch-map/source.ts
+++ b/src/fetch-map/source.ts
@@ -184,6 +184,11 @@ function getSourceOptions({
   };
 }
 
+/**
+ * Returns default tile resolution for dynamic tilesets, based on layer configuration
+ * Result is not applicable for static tilesets, for which tile resolution is defined
+ * by tilejson.
+ */
 function getDynamicTileResolution(
   spatialIndex?: SpatialIndex | null
 ): TileResolution {
--- a/src/fetch-map/source.ts
+++ b/src/fetch-map/source.ts
@@ -195,6 +195,11 @@ function getDynamicTileResolution(
   return DEFAULT_TILE_RESOLUTION;
 }
 
+/**
+ * State of `aggregationResLevel` in the UI and backend config is based on an assumption of
+ * 512x512px tiles. Because we may change tile resolution for performance goals, the
+ * `aggregationResLevel` passed to the deck.gl layer must be scaled with tile resolution.
+ */
 function scaleAggregationResLevel(
   aggregationResLevel: number,
   tileResolution: number
--- a/src/fetch-map/source.ts
+++ b/src/fetch-map/source.ts
@@ -1,0 +1,223 @@
+import {Dataset} from './types.js';
+import {SpatialIndex, SpatialIndexColumn} from '../constants.js';
+import {
+  QuerySourceOptions,
+  TableSourceOptions,
+  TilejsonResult,
+  TileResolution,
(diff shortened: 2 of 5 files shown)


src/fetch-map/source.ts
```diff
@@ -0,0 +1,223 @@
...
+    ...(queryParameters && {queryParameters}),
+  } as QuerySourceOptions;
+  const vectorOptions = {
+    spatialDataColumn,
+    ...(columns && {columns}),
+    ...(filters && {filters}),
+    ...(aggregationExp && {aggregationExp}),
+  } as VectorTableSourceOptions;
+
+  if (type === 'raster') {
```
zbigg, line 115: `SourceType` enum would be helpful

(yes, old `MapType`, which is actually type of source,, that's why i propose new name)

src/fetch-map/types.ts
```diff
@@ -205,3 +207,63 @@ export type GoogleBasemapProps = {
...
 };
+
+export type Dataset = {
+  id: string;
+  type: MapType;
+  source: string;
+  cache?: number;
+  connectionName: string;
+  geoColumn: string;
+  data: TilejsonResult | GeojsonResult | JsonResult;
```
zbigg, line 218: `GeojsonResult | JsonResult` not needed anymore?

zbigg: For future reuse Please split this type to "Dataset" (only things loaded from backend) and "LoadedDataset" (name todo) - with `data`, `cache` and other "fetchMap" specific goodies.

It would be good to use Dataset type that is not necessarily loaded

src/fetch-map/source.ts
```diff
@@ -0,0 +1,223 @@
...
+
+function getSpatialIndexFromGeoColumn(geoColumn: string) {
+  const spatialIndexToSearch = geoColumn.split(':')[0];
+
+  for (const index of Object.values(SpatialIndex)) {
+    if (SpatialIndexColumn[index].includes(spatialIndexToSearch)) {
+      return index;
+    }
+  }
+  return null;
```
zbigg, line 232: i know this code is old, but it it's overly complicated and makes no sense

what it does is simply
```const x = geoColumn.split(':')[0];
if (x === SpatialIndex.H3 || x === SpatialIndex.Quadbin) return spatialIndexToSearch
return null
```

zbigg (approved):
cosmetic comments, lgtm

src/fetch-map/source.ts
felixpalmer: Once I pull in the better type, this will use `MapType`: https://github.com/CartoDB/cloud-native/blob/main/workspace-www/src/features/maps/types.ts#L252

Is that what you want?

src/fetch-map/types.ts
felixpalmer: > GeojsonResult | JsonResult not needed anymore?

it would make sense to remove these also from `base-source`, @donmccurdy do you know any reason not to do this?

src/fetch-map/source.ts
```diff
@@ -0,0 +1,223 @@
...
+): TileResolution {
+  // TODO: Support increased tile size and resolution for dynamic H3 spatial indexes.
+  if (spatialIndex !== SpatialIndex.H3) {
+    return REDUCED_QUERIES_TILE_RESOLUTION;
+  }
+
+  return DEFAULT_TILE_RESOLUTION;
+}
+
+function scaleAggregationResLevel(
```
donmccurdy, line 208: Could you include the comments from the source files? I worry some of this code will be difficult to parse for future us.

```suggestion
/**
 * State of `aggregationResLevel` in the UI and backend config is based on an assumption of
 * 512x512px tiles. Because we may change tile resolution for performance goals, the
 * `aggregationResLevel` passed to the deck.gl layer must be scaled with tile resolution.
 */
function scaleAggregationResLevel(
```

src/fetch-map/source.ts
```diff
@@ -0,0 +1,223 @@
...
+  };
+}
+
+function getDynamicTileResolution(
+  spatialIndex?: SpatialIndex | null
+): TileResolution {
+  // TODO: Support increased tile size and resolution for dynamic H3 spatial indexes.
+  if (spatialIndex !== SpatialIndex.H3) {
+    return REDUCED_QUERIES_TILE_RESOLUTION;
+  }
```
donmccurdy, line 198:
```suggestion
/**
 * Returns default tile resolution for dynamic tilesets, based on layer configuration
 * Result is not applicable for static tilesets, for which tile resolution is defined
 * by tilejson.
 */
function getDynamicTileResolution(
  spatialIndex?: SpatialIndex | null
): TileResolution {
  // TODO: Support increased tile size and resolution for dynamic H3 spatial indexes.
  if (spatialIndex !== SpatialIndex.H3) {
    return REDUCED_QUERIES_TILE_RESOLUTION;
  }
```

src/fetch-map/types.ts
donmccurdy: I assume those are leftovers from document mode? OK with me to remove if so.

src/fetch-map/types.ts
```diff
@@ -205,3 +207,63 @@ export type GoogleBasemapProps = {
...
+  aggregationResLevel: number;
+  queryParameters: QueryParameters;
+};
+
+export enum AttributeType {
+  String = 'String',
+  Number = 'Number',
+  Timestamp = 'Timestamp',
+  Boolean = 'Boolean',
+}
```
donmccurdy: Minor – I'm mostly trying to use type unions (`'String' | 'Number' | ...`) for public-facing code, for consistency in the API, would that be OK here? Have made some exceptions, particularly for cases where the underlying string values are 'weird'.

src/constants-internal.ts
```diff
@@ -19,6 +19,9 @@ export const DEFAULT_MAX_LENGTH_URL = 7000;
 /** @privateRemarks Source: @deck.gl/carto */
 export const DEFAULT_TILE_RESOLUTION = 0.5;
 
+/** @internal */
+export const REDUCED_QUERIES_TILE_RESOLUTION = 1;
+
```
donmccurdy: The 'reduced queries' value is really a reference to an old feature flag, we should probably update `DEFAULT_TILE_RESOLUTION` to 1.0 the next time there's a breaking change opportunity. I've been using the `TODO(v0.5)` prefix for that purpose:

```suggestion
/**
 * @todo TODO(v0.5): Update DEFAULT_TILE_RESOLUTION and remove this constant.
 * @internal
 */
export const REDUCED_QUERIES_TILE_RESOLUTION = 1;

```

donmccurdy (approved):

PR #139, opened 2025-03-19
fetchMap: Client & server side filtering

felixpalmer:
Add filtering integration for server/client-side filtering

#### Change list

- Pass through filter to data source for server-side filters
- Expose filter for client side filtering
- Example of how to perfrom client-side filter

Screenshot 2025-03-19 at 16 29 53


examples/05-fetch-map/app.ts
```diff
@@ -43,13 +45,17 @@ const layerClasses: Record<LayerType, _ConstructorOf<Layer>> = {
...
       if (!LayerClass) {
         console.error(`No layer class found for type: ${type}`);
         return null;
       }
-      return new LayerClass(props);
+      const filterProps = filters && {
+        ...getDataFilterExtensionProps(filters),
+        extensions: [new DataFilterExtension({filterSize: 4})],
+      };
+      return new LayerClass({...props, ...filterProps});
```
donmccurdy, line 58: I'm not sure if we have validation to ensure `filters` will be falsy if there are no filters applied, might need to do `import {isEmptyObject} from './utils.js';` and check that as well? Unless we think there are no side effects of an unused DataFilterExtension...

donmccurdy (approved):

PR #153, opened 2025-04-15
 Export legendSettings

felixpalmer:

#### Change list

Allow access to the `legendSettings`


donmccurdy (approved):
Looks good thanks! Eventually I guess we will need to bring the type definitions over, but I'm not sure what the right timing for that would be – probably not until Builder integration.

PR #154, opened 2025-04-15
Legend example PoC

felixpalmer:

### Background

For discussion, regarding API for legends. I've marked as merging into https://github.com/CartoDB/carto-api-client/pull/153, but it doesn't actually need the top-level `legendSettings`.

### Demo

https://github.com/user-attachments/assets/49b0c08d-9a14-4d26-83ad-b4aa9e785acc

Example of how to use use the existing `props` that the `Layer` has to build a simple legend. On one hand, it is nice to be able to reuse the existing APIs, on the other I'm not sure if reaching deep into `props.data` to retrieve `widgetSource` and `tilestats` is very clean. Maybe we could wrap this better in some sort of `Widget`

### DIFF
--- a/examples/06-fetch-map/legend.ts
+++ b/examples/06-fetch-map/legend.ts
@@ -29,7 +29,7 @@ export function createLegend(layers: LayerDescriptor[]): HTMLElement {
     layerDiv.appendChild(nameDiv);
 
     const columnDiv = document.createElement('div');
-    columnDiv.textContent = `Data column: ${dataColumn}`;
+    columnDiv.textContent = `Color based on: ${dataColumn}`;
     layerDiv.appendChild(columnDiv);
 
     // Add min/max values from tilestats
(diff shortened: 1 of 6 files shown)


examples/06-fetch-map/legend.ts
```diff
@@ -0,0 +1,107 @@
...
+  wrapper.className = 'legend-wrapper';
+
+  const container = document.createElement('div');
+  container.className = 'legend-container';
+
+  layers.forEach((layer) => {
+    const widgetSource = layer.props.data?.widgetSource?.props;
+    if (!widgetSource) return;
+
+    const {columns, spatialDataColumn} = widgetSource;
```
donmccurdy, line 15: I don't think this matters for examples in this repository, but one thing to note for the future fetchMap/Builder integration: I don't think we're consistently initializing data sources (e.g. `h3TilesetSource`) in Builder with the correct spatialDataColumn, it looks like sometimes that is inferred later, and we use `vectorTilesetSource` without knowing (yet) whether it's geometry or quadbin or h3. So we might need a different approach than this in Builder.

examples/06-fetch-map/legend.ts
```diff
@@ -0,0 +1,107 @@
...
+      categoriesDiv.textContent = `Categories: ${tilestats.categories.length}`;
+      layerDiv.appendChild(categoriesDiv);
+    }
+
+    container.appendChild(layerDiv);
+  });
+
+  wrapper.appendChild(container);
+  return wrapper;
+}
```
donmccurdy, line 107: Optional – I think constructing the HTML programmatically is completely OK since we're trying to avoid  dependencies on frameworks here. If you did prefer something more ergonomic for this or future examples though, adding a lightweight HTML templater like [lit-html](https://www.npmjs.com/package/lit-html) would also be fine with me. (or others, mainly suggesting that because I've used it before and it's nearly dependency-free).

donmccurdy (approved):
>  Maybe we could wrap this better in some sort of Widget

I'm happy with how this example is structured and wouldn't insist on any change right now! But perhaps (whether now or later) the strategy could be to return a richer description of the Datasets? This could include direct access to the WidgetSource instances.

examples/06-fetch-map/legend.ts
felixpalmer: I like your philosophy of not tying api-client to any frontend framework, so I went in the spirit of that and avoiding adding a dependency. For such a small example I think it is OK to just use vanilla JS

felixpalmer:
> return a richer description of the Datasets

Could you eloborate? I'm not sure I understand what should be returning a richer description

donmccurdy:
For example:

```javascript
const { layers, datasets } = fetchMap({cartoMapId: '<map-id>'});
const dataset = datasets.find((dataset) => dataset.id === '<dataset-id>');
const { widgetSource } = dataset;
```

Perhaps I misunderstood "wrap this in some sort of Widget" – but I think that accessing a WidgetSource object should not require that the map has any widgets, legends, or even layers for that matter.

PR #156, opened 2025-04-16
chore(dev): Finish build before starting dev server

donmccurdy:


felixpalmer (approved):

package.json
```diff
@@ -39,7 +39,7 @@
   "scripts": {
     "build": "tsup",
     "build:watch": "tsup --watch",
-    "dev": "concurrently \"yarn build:watch\" \"vite --config examples/vite.config.ts --open\"",
+    "dev": "yarn build && concurrently --names tsup,vite \"yarn build:watch\" \"vite --config examples/vite.config.ts --open\"",
```
felixpalmer, line 42: what does `--names tsup,vite` do?

donmccurdy: Puts a prefix on the log output, so it's easier to see whether if an error is coming from the dev server (Vite) or the library builds (tsup). Not really necessary, but without it the prefixes would just be `[0]` and `[1]`.

CleanShot 2025-04-17 at 09 06 23@2x

PR #159, opened 2025-04-18
types needed to get new fetchMap working with deck.gl

felixpalmer:

#### Change list

- Types needed in deck.gl


src/index.ts
```diff
@@ -2,6 +2,8 @@ export * from './client.js';
 export * from './constants.js';
 export * from './deck/index.js';
 export * from './fetch-map/index.js';
+export {LayerType} from './fetch-map/layer-map.js';
+export {LayerDescriptor} from './fetch-map/parse-map.js';
```
donmccurdy: Could we export these from `./fetch-map/index.ts` if they're intended to be public? 

Also note that after https://github.com/CartoDB/carto-api-client/pull/160 they'd need to use `export type`.

felixpalmer: Done

donmccurdy:
@felixpalmer something's weird here, shouldn't these exports be redundant?

```diff
export * from './fetch-map/index.js';
+ export type {LayerDescriptor, LayerType} from './fetch-map/index.js';
```

felixpalmer:
> shouldn't these exports be redundant?

@donmccurdy not sure I follow, I thought the idea was that type exports should now be handled separately?

donmccurdy (approved):

donmccurdy:
Looks good after updates to export missing types used in public API, thanks!

PR #160, opened 2025-04-18
chore(build): Enable verbatimModuleSyntax=true

donmccurdy:
Based on https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library, I think it might be best to replace `esModuleInterop` and `allowSyntheticDefaultImports` with `verbatimModuleSyntax`. It's meant to help libraries ensure compatibility with consumers who might have different `tsconfig.json` settings.

Lots of one-time changes...

```diff
-import {MapType} from '../types.js';
+import type {MapType} from '../types.js';
```

... but the build will catch any missed type-only exports moving forward.

#### PR Dependency Tree

* **PR #160** 👈


jmgaya (approved):

jmgaya:
Great improvement! I didn't know about that option, but it looks like a really good addition 💯 

Take into account some files conflict with `main`, but I guess it's just about `import` statements, so it should be an easy one 💪

zbigg (approved):

PR #165, opened 2025-04-30
perf(fetchMap): Remove moment-timezone dependency

donmccurdy:
Bundle size of `@carto/api-client` accidentally increased to >1 MB because of the large `moment-timezone` dependency. The dependency is in [legacy mode and discouraged](https://www.npmjs.com/package/moment-timezone) anyway, I've replaced it with the web's [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) utility. This creates a small change in date formatting, but I think that's acceptable for our purposes.

I've also added a very basic regression test for the bundle size. Currently this detects if the inline bundle crosses a threshold — it doesn't check external dependencies! — but it would be enough to catch this case at least.

#### PR Dependency Tree

* **PR #165** 👈

### DIFF
--- a/src/fetch-map/utils.ts
+++ b/src/fetch-map/utils.ts
@@ -67,3 +67,21 @@ export function isRemoteCalculationSupported(dataset: Dataset) {
 
   return true;
 }
+
+const DATE_FORMATTER = new Intl.DateTimeFormat('en-US', {
+  year: '2-digit',
+  month: 'numeric',
+  day: 'numeric',
+  hour: 'numeric',
+  minute: 'numeric',
+  second: 'numeric',
+  timeZone: 'UTC',
+});
+
+export function formatDate(value: string | number | Date): string {
+  return DATE_FORMATTER.format(new Date(value));
+}
+
+export function formatTimestamp(value: string | number | Date): string {
+  return String(Math.floor(new Date(value).getTime() / 1000));
+}
--- a/test/bundle-size.test.ts
+++ b/test/bundle-size.test.ts
@@ -1,0 +1,16 @@
+import {expect, test} from 'vitest';
+import {readFile} from 'node:fs/promises';
+
+// Confirm bundle size is non-trivial (not a barrel file), but also not larger
+// than the allocated bundle size budget.
+const BUNDLE_SIZE_MIN = 10_000;
+const BUNDLE_SIZE_MAX = 400_000;
+
+test('bundle size', async () => {
+  const pkg = JSON.parse(await readFile('package.json', 'utf8'));
+  const entryPath = pkg.exports['.']['default']['default'];
+  const entry = await readFile(entryPath, 'utf8');
+
+  expect(entry.length).toBeGreaterThan(BUNDLE_SIZE_MIN);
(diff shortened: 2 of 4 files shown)


src/fetch-map/utils.ts
```diff
@@ -67,3 +67,21 @@ export function isRemoteCalculationSupported(dataset: Dataset) {
...
+
+const DATE_FORMATTER = new Intl.DateTimeFormat('en-US', {
+  year: '2-digit',
+  month: 'numeric',
+  day: 'numeric',
+  hour: 'numeric',
+  minute: 'numeric',
+  second: 'numeric',
+  timeZone: 'UTC',
+});
```
jmgaya, line 79: I won't worry too much right now, but I'm modifying how dates are being rendered in `cloud-native` through the new `DateRangePicker` component. Under the hood, we use `dayjs` for doing such (which doesn't mean we should be using it here), but I think at some point it could be interesting to use a similar format everywhere, or maybe let the client inject their own format while using `carto-api-client` 👍

test/bundle-size.test.ts
```diff
@@ -0,0 +1,16 @@
...
+const BUNDLE_SIZE_MAX = 400_000;
+
+test('bundle size', async () => {
+  const pkg = JSON.parse(await readFile('package.json', 'utf8'));
+  const entryPath = pkg.exports['.']['default']['default'];
+  const entry = await readFile(entryPath, 'utf8');
+
+  expect(entry.length).toBeGreaterThan(BUNDLE_SIZE_MIN);
+  expect(entry.length).toBeLessThan(BUNDLE_SIZE_MAX);
+});
```
jmgaya, line 16: Cleanest, yet simplest way to keep things under control, love it 😍 !

test/bundle-size.test.ts
```diff
@@ -0,0 +1,16 @@
+import {expect, test} from 'vitest';
```
jmgaya, line 1: Aren't these utilities already defined globally for every test? Or haven't we configured that in `carto-api-client`?

jmgaya (approved):
Great job, thanks for caring about these details 🫶

src/fetch-map/utils.ts
donmccurdy: Nice! Yes `dayjs` looks much more compact, I'd be OK with adding it here. In general I think it's good if we can keep the API client mostly light and unopinionated about views/formatting/UI ... but that goal does conflict a bit with the "reproduce a Builder map" purpose here, so I guess we'll need to work out a compromise at some point. :)

test/bundle-size.test.ts
donmccurdy: Didn't realize that was [a built-in option](https://vitest.dev/config/#globals) in Vitest! I'd just thought of it as being a Jest thing. Personally it's a little too much magic for me (TS auto-imports are near-instant in these smaller repositories anyway) but I'd be OK with adding it if that seems like a direction we want to go for consistency.

PR #163, opened 2025-04-29
fix(fetchMap): Support more clusterTile layer props

donmccurdy:
#### PR Dependency Tree

* **PR #163** 👈

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

PR #173, opened 2025-05-01
fix(fetchMap): Fix cluster layer default aggregation expression column

donmccurdy:
Follow-up for #163. I'd thought that using the first entry in `dataset.columns` would be the same thing as the first entry from `mapInstantiation.schema`, and preferred that because the API client didn't previously expose the schema, but it looks like those are not necessarily the same fields/order, leading to issues where the cluster layer defaults to the wrong field for its aggregation expression column.

Fixes cluster layer fetchMap rendering.

CleanShot 2025-05-01 at 12 27 38@2x

#### PR Dependency Tree

* **PR #173** 👈

### DIFF
--- a/src/fetch-map/layer-map.ts
+++ b/src/fetch-map/layer-map.ts
@@ -33,6 +33,7 @@ import type {
 } from './types.js';
 import type {ProviderType} from '../types.js';
 import {DEFAULT_AGGREGATION_EXP_ALIAS} from '../constants-internal.js';
+import type {SchemaField} from '../types-internal.js';
 
 const SCALE_FUNCS: Record<string, () => any> = {
   linear: scaleLinear,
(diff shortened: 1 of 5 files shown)


src/fetch-map/layer-map.ts
```diff
@@ -495,11 +496,11 @@ export function calculateClusterRadius(
 export function getDefaultAggregationExpColumnAliasForLayerType(
   layerType: LayerType,
   provider: ProviderType,
-  columns: string[]
+  schema: SchemaField[]
 ): string {
-  if (columns && layerType === 'clusterTile') {
+  if (schema && layerType === 'clusterTile') {
```
jmgaya, line 501: Ouch, I didn't remeber that, apologies for not highlighting it in the previous PR 🙏

jmgaya (approved):

src/sources/base-source.ts
```diff
@@ -85,5 +85,8 @@ export async function baseSource<UrlParameters extends Record<string, unknown>>(
   if (accessToken) {
     json.accessToken = accessToken;
   }
+  if (schema) {
+    json.schema = schema;
+  }
```
donmccurdy, line 90: /cc @felixpalmer just FYI, this does change the tilejson response slightly, including the schema fields from the map instantiation. I didn't see any other way to expose that, and we appear to need it in fetchMap(). But I'd be open to doing this differently in the future.

src/fetch-map/layer-map.ts
donmccurdy: I probably should have called it out as a copy/paste except for this change, sorry!

PR #174, opened 2025-05-09
WidgetRemoteSource: custom aggregations for category, timeseries models

zbigg:
Story: https://app.shortcut.com/cartoteam/story/486261

Just pass new parameters and update typings for `WidgetSource`s `getCategories` and `getTimeSeries`
 * `operation` can be also `custom`
 * `operationExp` needed if operation is `custom`

### DIFF
--- a/src/widget-sources/types.ts
+++ b/src/widget-sources/types.ts
@@ -33,7 +33,8 @@ export interface BaseRequestOptions {
 /** Options for {@link WidgetRemoteSource#getCategories}. */
 export interface CategoryRequestOptions extends BaseRequestOptions {
   column: string;
-  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
+  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum' | 'custom';
+  operationExp?: string;
   operationColumn?: string;
   /** Local only. */
   joinOperation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
@@ -131,7 +132,8 @@ export interface TimeSeriesRequestOptions extends BaseRequestOptions {
   column: string;
   stepSize: GroupDateType;
   stepMultiplier?: number;
-  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
+  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum' | 'custom';
+  operationExp?: string;
   operationColumn?: string;
   joinOperation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
   splitByCategory?: string;
--- a/src/widget-sources/types.ts
+++ b/src/widget-sources/types.ts
@@ -3,6 +3,7 @@ import type {
   TileResolution,
 } from '../sources/types.js';
 import type {
+  AggregationType,
   Filters,
   GroupDateType,
   SortColumnType,
@@ -33,7 +34,8 @@ export interface BaseRequestOptions {
 /** Options for {@link WidgetRemoteSource#getCategories}. */
 export interface CategoryRequestOptions extends BaseRequestOptions {
   column: string;
-  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum' | 'custom';
+  operation?: AggregationType
+  /** Remote only. Only valid if operation is 'custom' */
   operationExp?: string;
   operationColumn?: string;
   /** Local only. */
@@ -86,7 +88,8 @@ export interface FeaturesRequestOptions extends BaseRequestOptions {
 /** Options for {@link WidgetRemoteSource#getFormula}. */
 export interface FormulaRequestOptions extends BaseRequestOptions {
   column: string;
-  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum' | 'custom';
+  operation?: AggregationType
+  /** Remote only. Only valid if operation is 'custom' */
   operationExp?: string;
   joinOperation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
 }
@@ -132,7 +135,8 @@ export interface TimeSeriesRequestOptions extends BaseRequestOptions {
   column: string;
   stepSize: GroupDateType;
   stepMultiplier?: number;
-  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum' | 'custom';
+  operation?: AggregationType
+  /** Remote only. Only valid if operation is 'custom' */
   operationExp?: string;
   operationColumn?: string;
   joinOperation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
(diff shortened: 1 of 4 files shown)


src/widget-sources/widget-remote-source.ts
```diff
@@ -314,13 +323,18 @@ export abstract class WidgetRemoteSource<
...
+      operationExp,
       stepSize,
       stepMultiplier,
       splitByCategory,
       splitByCategoryLimit,
       splitByCategoryValues,
     } = params;
 
+    if (operation === 'custom') {
+      assert(operationExp, 'operationExp is required for custom operation');
```
jmgaya, line 335: It might come in handy to write a test that tests a `custom` operation without an `operationExp`, otherwise, we might remove this part, and the consumer could start using it wrongly

jmgaya (approved):
LGTM 🚢 !

src/widget-sources/widget-remote-source.ts
zbigg: > handy to write ... consumer could start using it wrongly

I think customer "might" is not enough excuse to add more lines to tests which have low chance of becoming real regression. In worst case backend will return 400.

This API is checked and many levels (typescipt, later backend), finally expectation that consumer uses API according to docs (we only suport this, if he doesn't we only help him with assetions like this).

And why this particular exceptioal case "have priviledge" of being covered by tests?
Are we asserting anything about `filters/spatialFilter` correctness of operation against enum?

This is thin integration layer API and really, it's enough that tests cover only major/important use cases and rest is just left for typescript/user level testing.

Other spectrum is that we use property testing and generate all possible error input classes, but ... that's not we want to do _now_ in this project circumstances.

jmgaya: The presence of `custom` + `operationExp` is not checked by TypeScript, but it's something your code assumes (with this `if` statement), isn't it?

I consider that a test that checks an `if` branch is worth it, and I wrote it to prevent not only the consumer from using it wrongly, but us removing it and not receiving any immediate feedback about something that is clearly wrong 🤷

src/widget-sources/types.ts
```diff
@@ -33,7 +33,8 @@ export interface BaseRequestOptions {
 /** Options for {@link WidgetRemoteSource#getCategories}. */
 export interface CategoryRequestOptions extends BaseRequestOptions {
   column: string;
-  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
+  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum' | 'custom';
```
donmccurdy: Optionally, could use `AggregationType` here since it now includes all values from the type union.

Also, do you mind including a `/** Remote only. */` comment similar to the one on `joinOperation`? I know users probably won't see it but this does help when I need to remember which operations we support in remote vs. tileset widgets. 🙏🏻

src/widget-sources/types.ts
```diff
@@ -131,7 +132,8 @@ export interface TimeSeriesRequestOptions extends BaseRequestOptions {
   column: string;
   stepSize: GroupDateType;
   stepMultiplier?: number;
-  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
+  operation?: 'count' | 'avg' | 'min' | 'max' | 'sum' | 'custom';
```
donmccurdy: See above.

src/widget-sources/widget-remote-source.ts
donmccurdy: My preferences, in order of priority:

1. Use TypeScript to enforce valid input when practical, without hurting readability and maintainability (in this case, probably too messy...)
2. Rely on the API to report incorrect usage with HTTP 4XX errors
3. Validate input in the API client only if (a) incorrect usage is not adequately handled by the API, or (b) the client-side validation has some extra value, like simplifying the code following the assertion

I think that _if_ the client-side validation is present (maybe it isn't necessary here?) then ideally it would also be covered by tests, but as always there's some balance of time/cost/benefit and I'm happy either way in this case.

Very possibly there are other validation checks in the API client that don't follow 1–3, I haven't thought so much about it, and some checks were probably inherited from C4R.

donmccurdy (approved):

src/widget-sources/widget-remote-source.ts
jmgaya: Yup, I agree with you both that there's always a line between "pragmatism" and "dogmatism", and probably depending on the scenario, it might be worth having it covered or not 💯 

In this case, the API will probably complain about it, and if the Documentation warns the user, we already have a "protection layer" somehow, so it might not be totally worth it. I didn't know the API was already handling this properly neither, so it was more like a "Hey, if we remove this, we could let the user make 'dirty things', so let's protect ourselves", but yeah, glad to had this conversation with you both, thanks a lot 🙇 !

src/widget-sources/types.ts
zbigg: I've just realized that `operation` is mandatory, at least for category/timeseries/formula`

Can we remove `?` ... i think we should, just asking for opinions

src/widget-sources/widget-remote-source.ts
zbigg: What i wanted to say is that "assertions" in language like JS, are convenience tools that don't change behavior of program _significantly_ (even if they are in IF).

Removal of it would only make "devx" slighly worse and only if he's in trouble.

I can remove it and e2e behavior will be same - error (400 or 500 in backend depending on how strict it is - it is, they use zod-based proper schape validation and it would be caught).

That's why in this case, i believe that another test case for this would be of "very low utility".

Contrast it with this PR: https://github.com/CartoDB/carto-api-client/pull/178 which as a whole basically new _critical_ "exceptional case", that _if_ removed or broken will also break widgets. Test case for this change is certainly valuable and i would vote for it :)

zbigg: Anyway, i can try to replicate `AggOperationParams` (look at maps-api/src/lib/query-models/models/timeseries.ts) here, so we have proper typing.

src/widget-sources/types.ts
donmccurdy: @zbigg for tileset widgets it is optional and defaults to 'count' ...

https://github.com/CartoDB/carto-api-client/blob/fd34134af9dfc48b88dae6f4e0ea26f484b273cf/src/widget-sources/widget-tileset-source-impl.ts#L121-L123

... I'm not sure if changing that would break things, or whether there are legacy reasons vs. just an unplanned difference. So it might be safer to default to 'count' everywhere. I'm OK with either but would probably want to make this a v0.5 -> v0.6 bump to be safe if we're making the property required for tilesets.

src/widget-sources/widget-remote-source.ts
donmccurdy: @zbigg as you prefer, approved already so you can merge this whenever you feel it's ready!

donmccurdy: Improving test coverage for #178 in https://github.com/CartoDB/carto-api-client/pull/180. :)

PR #176, opened 2025-05-12
chore(deps): Unpin h3-js

### LINKED ISSUE #175: h3 dependency version
I just tried to pull the latest api-client into deck.gl and get this change in `yarn.lock`

Image

Is there a reason we are pinning to a precise version in `package.json`? In deck.gl we use "^4.1.0". As `carto-api-client` is aiming to be consumed as a library would it not be better to have a looser version requirement to minimize impact on consumers?

donmccurdy:
- Fixes #175

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@types/geojson": "^7946.0.16",
     "d3-format": "^3.1.0",
     "d3-scale": "^4.0.2",
-    "h3-js": "^4.2.1",
+    "h3-js": "^4.1.0",
     "quadbin": "^0.4.1-alpha.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@types/geojson": "^7946.0.16",
     "d3-format": "^3.1.0",
     "d3-scale": "^4.0.2",
-    "h3-js": "4.2.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -67,7 +67,7 @@
     "@types/geojson": "^7946.0.16",
     "d3-format": "^3.1.0",
     "d3-scale": "^4.0.2",
-    "h3-js": "4.2.1",
+    "h3-js": "^4.2.1",
```
jmgaya: Outside of this PR scope, but it might be worth creating a ticket in `cloud-native`, so we align this dependency there, instead of bundling both versions 👍

jmgaya (approved):

package.json
felixpalmer: deck.gl is using `4.1.0`, so can we do this instead? Otherwise we will be packaging 3 versions of h3 (cloud-native is on `3.7.2`)


```suggestion
    "h3-js": "^4.1.0",
```

felixpalmer (changes requested):

felixpalmer (approved):

package.json
donmccurdy: Changed to ^4.1.0 here, and continued in:

- https://github.com/CartoDB/carto-kepler/pull/196

PR #179, opened 2025-05-14
Expose legend info from fetchMap

felixpalmer:
Proposal for exposing the info needed to build legends. For each layer, `layer.props.scales` contains a map for all the legends to be displayed, containing the following keys:

- `field`: `{name: string, type: string}`
- `domain`: `(number | string)[]`
- `range`: `string[]`
- `type`: `string` // Scale type

For example

```ts
{
  "fillColor": {
    "field": {
      "name": "population",
      "type": "integer"
    },
    "domain": [
      1,
      10,
      100,
      1000,
      10000,
      null
    ],
    "range": [
      "#f7feae",
      "#b7e6a5",
      "#7ccba2",
      "#46aea0",
      "#089099",
      "#00718b"
    ],
    "type": "custom"
  }
}
```


zbigg (commented):
I know it's PoC, but ...

> layer.props.scales

Please move it to `layer.scales` (to `LayerDescriptor` type)

`Layer.props` is meant to be passed directly to layer and yes, `scales` is probably harmless but it shouldn't be passed randomly to layer

examples/06-fetch-map/legend.ts
```diff
@@ -1,107 +1,106 @@
...
+      if (!dataColumn) {
+        const color = layer.props.getFillColor;
+        const rgb = `rgb(${color[0]}, ${color[1]}, ${color[2]})`;
+        layerDiv.appendChild(createColorSwatch(rgb, layer.props.cartoLabel));
+      } else {
+        const {domain, range, type} = scaleInfo;
+        if (
+          type === 'ordinal' ||
+          type === 'point' ||
+          typeof domain[0] === 'string'
```
donmccurdy: Do the if/else blocks here need to depend on type checks on the domain values, or would checks purely on the scale type be enough?

examples/06-fetch-map/legend.ts
```diff
@@ -1,107 +1,106 @@
...
+            const labelText =
+              end === null || end === undefined || end === Infinity
+                ? `>${start.toFixed(1)}`
+                : `${start.toFixed(1)} – ${end.toFixed(1)}`;
+            layerDiv.appendChild(createColorSwatch(range[i], labelText));
+          }
+        } else if (
+          typeof domain[0] === 'number' &&
+          typeof domain[1] === 'number' &&
+          range.length > 1
```
donmccurdy: Similarly, just wondering if `type === 'linear'` would have the same effect here?

src/fetch-map/layer-map.ts
```diff
@@ -39,7 +39,24 @@ import type {ProviderType} from '../types.js';
...
+export type SCALE_TYPE =
+  | 'linear'
+  | 'ordinal'
+  | 'log'
+  | 'point'
+  | 'quantile'
+  | 'quantize'
+  | 'sqrt'
+  | 'custom'
+  | 'identity';
```
donmccurdy: This list is slightly different than the one I see in Kepler (missing "logarithmic", not "log"), is there a conflicting source-of-truth for the types somewhere?

src/fetch-map/parse-map.ts
```diff
@@ -29,13 +29,22 @@ import type {
...
+  domain: (string | number)[];
+  range: string[];
+  type: SCALE_TYPE;
+};
+
 export type LayerDescriptor = {
   type: LayerType;
   props: Record<string, any>;
   filters?: Filters;
+  scales: Record<string, Scale>;
```
donmccurdy: Can/should this be `Record<ScaleKey, Scale>` instead?

src/fetch-map/parse-map.ts
```diff
@@ -29,13 +29,22 @@ import type {
   VisConfig,
   MapConfigLayer,
   Dataset,
+  VisualChannelField,
 } from './types.js';
 import {isRemoteCalculationSupported} from './utils.js';
 
+export type Scale = {
+  field: VisualChannelField;
+  domain: (string | number)[];
```
donmccurdy: Safe to make this `string[] | number[]` instead?

src/fetch-map/parse-map.ts
```diff
@@ -29,13 +29,22 @@ import type {
...
+  range: string[];
```
donmccurdy: Surprised the type of `range` is `string[]`, never numeric, is this intended?

src/fetch-map/layer-map.ts
```diff
@@ -39,7 +39,24 @@ import type {ProviderType} from '../types.js';
...
 
-const SCALE_FUNCS: Record<string, () => any> = {
+export type D3Scale = {
+  domain: (d?: any) => any[];
+  range: (d?: any) => any[];
+  unknown?: (d?: string) => any;
+} & ((d: any) => any);
+type D3ScaleFactory = () => D3Scale;
+
+export type SCALE_TYPE =
```
donmccurdy: For consistency let's use `ScaleType` (title case) to refer to a type union or enum (intentionally diverging from Kepler).

donmccurdy (approved):

examples/06-fetch-map/legend.ts
felixpalmer: Honestly I'm not sure, as the API isn't specified anywhere. The conditions are the best I could come up with by trying out some test maps. As discussed in the meeting, the implementation should be tested against more maps and at that point we can refine

examples/06-fetch-map/legend.ts
felixpalmer: Done, I also added `quantize`

src/fetch-map/parse-map.ts
felixpalmer: No, good catch!

src/fetch-map/layer-map.ts
felixpalmer: Don't know the API well enough, when a logarithmic scale is selected in Builder the type ends up being `'custom'`

PR #187, opened 2025-05-21
chore(widgets): add AggregationTypes enum-like definition

zbigg:
Widgets: `AggregationTypes` enum-like container for all `AggregationType` constants.

* [x] `AggregationTypes`
* [x] use constants from `AggregationTypes` in this package


src/types.ts
```diff
@@ -108,6 +108,20 @@ export type AggregationType =
...
+ * @privateRemarks Source: @carto/constants
+ */
+export const AggregationTypes: Record<AggregationType, AggregationType> = {
+  count: 'count' as const,
+  avg: 'avg' as const,
+  min: 'min' as const,
+  max: 'max' as const,
+  sum: 'sum' as const,
+  custom: 'custom' as const,
+} as const;
```
zbigg: Re case
* `COUNT` - we don't do caps anymore in new code?
* `count` - idk, but consistency with enum values feels good and flexible, not sure what are problems with it, i just happen to like it
* `Count` - yet another convention, actually i would go for it, if we insist that bare lower-case is problematic
     ... ok, it's consistent with backend
     
     
@CartoDB/app-dev-builder-team thoughts?

donmccurdy: I think I would prefer either (a) `EnumName.VALUE` or (c) `EnumName.Value`. (a) is more common in public TypeScript code, and matches other enums currently in this repository:

https://github.com/CartoDB/carto-api-client/blob/main/src/constants.ts

... but (c) matches CARTO conventions, I'd be fine with changing this repo to follow that instead.

In either case – let's move this from `types.ts` to `constants.ts` since it's a runtime constant, similar to `Filter` (type) vs. `FilterType` (enum).

zbigg: Moved to constants and changed symbols to CamelCase

donmccurdy (approved):

PR #189, opened 2025-05-23
feat(widgets): Distribute worker script as standalone JS bundle.

zbigg:
Distribute worker script in bundled version.

We expose worker as bundled JS script in `worker-compat` entry.

Example usage:

```ts
// vite
import cartoApiWorkerUrl from '@carto/api-client/worker-compat?url'
new WidgetTilesetSource({..., workerUrl: cartoApiWorkerUrl})

// manual asset managent
cp node_modules/@carto/api-client/build/worker-compat.js dist/static/carto-api-client-worker-compat.js
new WidgetTilesetSource({..., workerUrl: '/static/carto-api-client-worker-compat.js'})
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,9 @@
     "./worker": {
       "types": "./build/worker.d.ts",
       "default": "./build/worker.js"
+    },
+    "./worker.global": {
+      "default": "./build/worker.global.js"
     }
   },
   "browserslist": [
--- a/package.json
+++ b/package.json
@@ -31,8 +31,8 @@
       "types": "./build/worker.d.ts",
       "default": "./build/worker.js"
     },
-    "./worker.global": {
-      "default": "./build/worker.global.js"
+    "./worker-compat": {
+      "default": "./build/worker-compat.js"
     }
   },
   "browserslist": [
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "homepage": "https://github.com/CartoDB/carto-api-client#readme",
   "author": "Don McCurdy <donmccurdy@carto.com>",
   "packageManager": "yarn@4.3.1",
-  "version": "0.5.6-alpha.bundle.2",
+  "version": "0.5.6-alpha.bundle.3",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -30,6 +30,9 @@
     "./worker": {
       "types": "./build/worker.d.ts",
       "default": "./build/worker.js"
+    },
+    "./worker.global": {
```
zbigg: i have no idea what would be proper name for this 🤷 , `worker.global.js` is just something that `tsup` builds by default if target is `iife`

tsup.config.ts
```diff
@@ -25,4 +25,11 @@ export default defineConfig([
     entry: {'api-client': 'src/index.ts'},
     shims: true, // replace 'import.meta' references
   },
+  {
+    ...commonConfig,
+    format: 'iife',
```
zbigg, line 30: and i have no idea why we need `iife` and if it's best solution - this is just first way to convince `tsup` to _bundle_ all required deps into one file

note: `build/worker.js` contains normal `import` clauses so it works only in context where `imports` do work ... and they don't work in scripts loaded from `blob:` and/or in sandboxed iframes

package.json
donmccurdy: Perhaps `@carto/api-client/worker-compat`?

I've seen this `*-compat` naming used for "compatibility-mode" builds of WASM libraries, using inline base64 instead of binaries. I think the same intent describes what we're doing here, and this way we could change the implementation (if needed) without renaming.

tsup.config.ts
donmccurdy: Huh! Interesting – I'm OK with going this direction for now, if we find another way to bundle the imports later we can always change it back. 

Not necessary in this PR, but later we could add a short unit test that checks to make sure there are no imports in the `./worker-compat` script.

donmccurdy (approved):

package.json
zbigg: Done,

Bundle is `build/worker-compat.js` and exists in `exports` as `worker-compat`.

PR #192, opened 2025-05-27
perf(rasters): Improve raster widget performance for spatial filters covering many cells

donmccurdy:
To compute raster widget results within a spatial filter, we currently compute a [covering](http://s2geometry.io/devguide/examples/coverings.html) for the spatial filter, composed of quadbin cells at the same resolution as the raster tile pixels. When the raster pixels are small, or the spatial filter is large, the number of cells in the spatial filter covering set may be very large. Previously we used a JavaScript [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instance to check which pixels are contained by the spatial filter, but Sets are [limited to about ~16M features](https://stackoverflow.com/a/72809580/1314762) in today's JavaScript runtimes, and we may occasionally create a spatial filter covering >4000x4000 pixels.

The problem can be addressed by one — or more — of the following options:

1. Instead of choosing a resolution for the spatial filter based on the raster tile resolution, choose an independent resolution based on the size of the spatial filter region, and query against that. _Results from raster widgets will be approximate._
2. Replace or modify the `geometryToCells` implementation in quadbin-js to return covering cells at mixed resolution. If a large block of the spatial filter's interior can be covered by a much larger cell, 'contains' queries with the parent cell will be faster than iterating over its descendants, and cell counts will be much lower.
3. Avoid using JavaScript Set classes, and implement a custom lookup function

I think we should strongly consider doing (1) or (2), but (1) will require some product discussion and (2) sounds more complex than I'm prepared to implement right now. In the interest of unblocking further development and QA on the raster widgets feature, I've implemented (3) instead, which may complement the other two solutions anyway.

Rather than using Set, we use a custom HashSet implementation based on [MurmurHash2](https://en.wikipedia.org/wiki/MurmurHash). This avoids a hard limit of ~16M cells in the spatial filter, and otherwise has very similar performance compared to Set. I've hashed the entire quadbin key, but some further optimizations might be possible knowing that all cells are at the same resolution.

- Shortcut: [sc-489939]

#### PR Dependency Tree

* **PR #192** 👈

### DIFF
--- a/src/filters/tileFeaturesRaster.ts
+++ b/src/filters/tileFeaturesRaster.ts
@@ -11,6 +11,7 @@ import type {
   RasterMetadataBand,
   SpatialDataType,
 } from '../sources/types.js';
+import {CellSet} from '../utils/CellSet.js';
 
 export type TileFeaturesRasterOptions = {
   tiles: RasterTile[];
@@ -45,7 +46,7 @@ export function tileFeaturesRaster({
 
   // Compute covering cells for the spatial filter, at same resolution as the
   // raster pixels, to be used as a mask.
-  const spatialFilterCells = new Set(
+  const spatialFilterCells = new CellSet(
     geometryToCells(options.spatialFilter, cellResolution)
   );
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,4 +28,5 @@ export * from './filters/index.js';
 export * from './operations/index.js';
 export * from './utils/makeIntervalComplete.js';
 export * from './utils/transformToTileCoords.js';
+export * from './utils/CellSet.js';
 export * from './fetch-map/source.js';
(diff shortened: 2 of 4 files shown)


donmccurdy:
I made an initial attempt at (2) using `@mapbox/tile-cover`, but it creates the coverage set by first computing all cells at the highest resolution (as we do now), and then post-processing that list to combine cells into lower-resolution parents...

https://github.com/mapbox/tile-cover/blob/fecd2cfad1380e6e77600d0664f41b900eb1f290/index.js#L89-L101

... which reduces the cell count as expected, but is very slow and doesn't help with performance. We'd need a faster way to do that.

felixpalmer:
Hmm, tricky. Looking at the tile-cover code, the `polygonCover` function is the issue. Perhaps this block:

https://github.com/mapbox/tile-cover/blob/fecd2cfad1380e6e77600d0664f41b900eb1f290/index.js#L169-L178

could be modified to detect when the ranges the for loops are covering are large, and instead filling with larger tiles. It wouldn't work perfectly in all cases, but could be a simple way to improve the performance

jmgaya (approved):

PR #193, opened 2025-05-29
feat(sources): allow passing tags to baseSource and query methods

juandjara:
In the context of [this initiative](https://app.shortcut.com/cartoteam/write/IkRvYyI6I3V1aWQgIjY4MWExNDI3LTU5MzktNGE5ZC1hNjUzLTczYmI1MzZjZDZhMyI=)

We need a way to allow passing a set of tags to the request url of bath maps API calls and SQL API calls. I have drafted this initial implementation to be as simple as possible to not modify much the already existing behaviour of carto-api-client but allow for the extensibility this initiative requires.

I'll set this to draft until I have tested it on cloud native linking locally, but I'll leave this here in the meantime so we can comment the implementation

### DIFF
--- a/src/sources/types.ts
+++ b/src/sources/types.ts
@@ -76,7 +76,7 @@ export type SourceOptionalOptions = {
    */
   localCache?: LocalCacheOptions;
 
-  /** Additional tags appended to the request url that can be used to track calls for security purposes. */
+  /** Additional tags appended to HTTP request urls, available for analytics and audits. */
   tags?: Record<string, string>;
 };
 
--- a/src/sources/types.ts
+++ b/src/sources/types.ts
@@ -75,6 +75,9 @@ export type SourceOptionalOptions = {
    * By default, local in-memory caching is enabled.
    */
   localCache?: LocalCacheOptions;
+
+  /** Additional tags appended to the request url that can be used to track calls for security purposes. */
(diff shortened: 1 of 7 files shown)


src/sources/types.ts
```diff
@@ -75,6 +75,9 @@ export type SourceOptionalOptions = {
    * By default, local in-memory caching is enabled.
    */
   localCache?: LocalCacheOptions;
+
+  /** Additional tags appended to the request url that can be used to track calls for security purposes. */
+  tags?: Record<string, string>;
```
donmccurdy: I perhaps wouldn't say "security purposes" here, maybe something like:

```
Additional tags appended to HTTP requests, available for analytics and audits.

```

juandjara: Ah yes, "audits" was another word I was considering. Let's go with that

juandjara: Done

donmccurdy (approved):
My one concern is that maybe `additionalParameters` should be named as `internalParameters` and tagged with...

```js
/**
 * @internal
 * @experimental
 */
```

See longer explanation in Slack. I'll go ahead and approve the PR, if others disagree about the name then feel free to ignore this comment and submit as-is. Thank you!

PR #194, opened 2025-05-30
feat(widgets): Add othersThreshold to WidgetSource.getCategories

zbigg:
Support for calculation of "others/rest" `WidgetSource.getCategories`.

New args:
 * `othersThreshold ` whether widget source should calculate aggregations of "others" - aggregation of values not fitting first N categories
   * supported by "remote widgets" i.e: `WidgetRemoteSource`
   * supported by local `WidgetTilesetSource`
 * `rawResult` - returns response in as received from backend/worker, so in sligtly different shape: 
 
 ```
 {
   rows: [...], 
   metatada: {others: 123 // optional }
}
```

If `rawResult` is not specified, others is returned as entry named `_carto_others` for backwards compatibility.

### CODE REVIEW

File: src/widget-sources/types.ts [hunk 1]
```diff
@@ -59,6 +59,8 @@ export interface CategoryRequestOptions extends BaseRequestOptions {
   operationColumn?: string;
   /** Local only. */
   joinOperation?: 'count' | 'avg' | 'min' | 'max' | 'sum';
+  /** Calculate `_carto_others` category for all categories after first N (N is threshold). Remote only. */
+  othersThreshold?: boolean;
```
jmgaya: If `othersThreshold` is a `boolean`, how can we specify the number `N` for telling the Backend at which **threshold** we should calculate the "rest"?

zbigg: it's mistake! already fixed, it's number now

src/widget-sources/types.ts
```diff
@@ -206,7 +208,7 @@ export type FeaturesResponse = {rows: Record<string, unknown>[]};
 export type FormulaResponse = {value: number | null};
 
 /** Response from {@link WidgetRemoteSource#getCategories}. */
-export type CategoryResponse = {name: string; value: number}[];
+export type CategoryResponse = {name: string | null; value: number}[];
```
zbigg: Note, current category model supports SQL `NULL` and return it as JSON `null`, so `name` can be `null`.

src/operations/groupBy.ts
```diff
@@ -48,12 +51,27 @@ export function groupValuesByColumn({
...
+    return [];
   }
 
-  return [];
+  const allCategories = Array.from(groups).map(([name, value]) => ({
+    name,
+    value: targetOperation(value),
+  }));
+
+  allCategories.sort((a, b) => b.value - a.value);
```
jmgaya: If inmutability is preferred (I guess so?) we can chain the previous `.map` function with the `.sort` one

src/operations/groupBy.ts
```diff
@@ -48,12 +51,27 @@ export function groupValuesByColumn({
...
+
+  if (othersThreshold && allCategories.length > othersThreshold) {
+    const otherNames = allCategories
+      .map((entry) => entry.name)
+      .slice(othersThreshold);
+    const otherValue = otherNames.flatMap((name) => groups.get(name));
```
jmgaya: As `othersThreshold` is usually < 20, there shouldn't be any performance impact, but you're iterating twice when computing `otherNames` and `otherValue`, so feel free to do something like:

```
const others = allCategories.slice(othersThreshold);
const otherValues = others.flatMap(({ name }) => groups.get(name));
```

jmgaya (approved):
Thanks for cross-implementing this! 🚀

donmccurdy (approved):

PR #201, opened 2025-06-03
perf(rasters): Use per-tile spatial filter coverage for filtering cells

donmccurdy:
Refactors how we're checking raster cells against the current spatial filter. Instead of computing one large lookup table for the entire spatial filter, we check each tile to see whether it is fully outside, fully inside, or partially intersects the spatial filter. If the tile partially intersects, then we compute a smaller lookup table just for that tile and proceed with filtering its cells. Technical to follow, in Slack.

This change mostly eliminates the cost of filtering cells against the spatial filter, because (a) most tiles skip the filter, and (b) queries on the smaller lookup tables are faster. For particularly very complex spatial filters (global rivers?) there's still a bottleneck, not measured here, but such filters are not supported in Builder and are not currently a concern.

Results: For large datasets at zoom=3 and resolution=14, the change reduces processing time from 25s to 10.5s on my macbook pro. The remaining time is mostly unrelated to the spatial filter, but comes from converting the binary features into FeatureData. Leaving that part for a separate PR and discussion.

#### PR Dependency Tree

* **PR #201** 👈

### DIFF
--- a/src/filters/tileFeaturesRaster.ts
+++ b/src/filters/tileFeaturesRaster.ts
@@ -14,7 +14,7 @@ import type {
 } from '../sources/types.js';
 import booleanWithin from '@turf/boolean-within';
 import intersect from '@turf/intersect';
-import { feature, featureCollection } from '@turf/helpers';
+import {feature, featureCollection} from '@turf/helpers';
 
 export type TileFeaturesRasterOptions = {
   tiles: RasterTile[];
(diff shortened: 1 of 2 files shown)


src/filters/tileFeaturesRaster.ts
```diff
@@ -44,23 +47,34 @@ export function tileFeaturesRaster({
...
   for (const tile of tiles as Required<RasterTile>[]) {
     const parent = tile.index.q;
 
-    const children = cellToChildrenSorted(parent, cellResolution);
+    // If tile is partially overlapping with the spatial filter, compute a quadbin covering for the
+    // spatial filter + tile intersection, at cell resolution. If tile is fully inside or outside
+    // the spatial filter, computing a covering can be skipped. Avoid computing a quadbin covering
+    // for the _entire_ spatial filter, which may contain far too many cells (e.g. at zoom=3 and
+    // resolution=14, we have ~18,000,000 cells in the viewport.)
+    const tilePolygon = cellToBoundary(parent);
```
zbigg, line 60: Great optimization

zbigg (approved):

PR #203, opened 2025-06-04
feat(widgets): orderBy support in WidgetSource.getCategory

zbigg:
Support for `orderBy` for in `WidgetSource.getCategory`


src/widget-sources/types.ts
```diff
@@ -31,6 +31,12 @@ export interface BaseRequestOptions {
   filterOwner?: string;
 }
 
+export type CategoryOrderBy =
+  | 'frequency_asc' // sort by aggregate value ascending, then by name ascending
+  | 'frequency_desc' // sort by aggregate value descending, then by name ascending
+  | 'alphabetical_asc' // sort by category name ascending, then by value descending
+  | 'alphabetical_desc'; // sort by category name descending, then by value descending
```
donmccurdy, line 38: On the table widget we already have these options:

```
  sortBy?: string; // column name
  sortDirection?: SortDirection;
```

For consistency, perhaps we could name the params this way for the category widget?

```
  sortBy?: 'name' | 'value'
  sortDirection?: SortDirection
```

I'm not sure "frequency" is the right term, since we might be sorting max or min aggregation for example.

donmccurdy (approved):
Pre-approving the PR for whichever param names end up decided. 😅

PR #204, opened 2025-06-05
fix(tileset,raster): Support widget calculations without spatial filter

donmccurdy:
WidgetTilesetSource and WidgetRasterSource previously required a spatial filter to return widget results. With this update the spatial filter is no longer required. Omitting the spatial filter is preferred when the entire map is visible, to skip unnecessary checks against an all-inclusive bounding box. 

To support this change for all local layer types (geo, quadbin, h3, raster), and for easier understanding of the differences between them, I've refactored the intersection code into a common file (`tileIntersection.ts`). The general approach remains the same as #201, now applied to other layer types. 

H3 is a bit of a special case; child cells are not necessarily within the parent's boundary. For now I've kept a global covering set (not per-tile) for H3, to avoid any changes to E2E results, but for future reference we probably do have other options with [`polygonToCellsExperimental`](https://h3geo.org/docs/api/regions#polygontocellsexperimental).
 
Fixes [sc-491354].

#### PR Dependency Tree

* **PR #204** 👈

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "homepage": "https://github.com/CartoDB/carto-api-client#readme",
   "author": "Don McCurdy <donmccurdy@carto.com>",
   "packageManager": "yarn@4.3.1",
-  "version": "0.5.7-alpha.6",
+  "version": "0.5.7-alpha-optional-spatial-filter.0",
   "license": "MIT",
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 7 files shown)


zbigg (approved):

donmccurdy:
~~Failing E2E tests on tileset widgets while testing this PR – will need to debug a bit before merging.~~ Fixed.

PR #216, opened 2025-06-30
fix(constants): Export RasterBandColorinterp

donmccurdy:
#### PR Dependency Tree

* **PR #216** 👈

### DIFF
--- a/src/sources/constants.ts
+++ b/src/sources/constants.ts
@@ -1,0 +1,8 @@
+export enum RasterBandColorinterp {
+  Gray = 'gray',
+  Red = 'red',
+  Green = 'green',
+  Blue = 'blue',
+  Alpha = 'alpha',
+  Palette = 'palette',
+}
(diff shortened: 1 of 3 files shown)


src/sources/constants.ts
```diff
@@ -0,0 +1,8 @@
+export enum RasterBandColorinterp {
```
jmgaya, line 1: I might be wrong, but shouldn't it be `RasterBandColorInterp` (notice the **I** in `Interp`)

jmgaya (approved):

src/sources/constants.ts
donmccurdy: Ideally probably yes – bit of a weird case because we have these raster metadata fields we don't control from the API:

https://github.com/CartoDB/carto-api-client/blob/89a3a1459fd61e1f2b4bee9444ea26e62e5a0d89/src/sources/types.ts#L384-L415

But for now just want this to be a drop-in replacement with a Builder PR.

PR #220, opened 2025-07-07
feat(widgets): Add experimental getExtent() API

donmccurdy:
Adds _experimental_ `getExtent()` method to all WidgetSource classes. The API is currently experimental, and may receive breaking changes or deprecations outside of semver-major release versions.

Example:

```js
const { ...data, widgetSource } = await vectorTableSource(...);
const { bbox } = await widgetSource.getExtent();
```

As in turf.js, the bbox is given in [xmin, ymin, xmax, ymax] order.

[sc-497971]

#### PR Dependency Tree

* **PR #220** 👈

### DIFF
--- a/examples/01-basic/app.ts
+++ b/examples/01-basic/app.ts
@@ -81,7 +81,11 @@ function updateLayers() {
 
   deck.setProps({layers: [layer]});
   data
-    .then(({attribution}) => {
+    .then(({attribution, widgetSource}) => {
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      widgetSource.getExtent({}).then((response) => {
+        console.log('getExtent', response);
+      });
       document.querySelector('#footer')!.innerHTML = attribution;
     })
     .catch((error) => {
--- a/examples/01-basic/app.ts
+++ b/examples/01-basic/app.ts
@@ -81,11 +81,7 @@ function updateLayers() {
 
   deck.setProps({layers: [layer]});
   data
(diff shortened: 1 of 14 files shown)


examples/01-basic/app.ts
```diff
@@ -81,7 +81,11 @@ function updateLayers() {
 
   deck.setProps({layers: [layer]});
   data
-    .then(({attribution}) => {
+    .then(({attribution, widgetSource}) => {
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      widgetSource.getExtent({}).then((response) => {
+        console.log('getExtent', response);
```
zbigg: improvement idea:
I think it would be better to example, if we would use extent for something "tangible", like setting initial view state?

 We can do it after initial `updateSources` call, like that:

```
updateSources();
data.then(( async {widgetSource}) => {
    cosnt extent = await widgetSource.getExtent({})
    console.log("...")
    // but we can also set viewState in deck instance to showcase how to use this extent for what user usually do - see whole data
    
    deck.setProps({viewState: ...})
})
```

zbigg (approved):
looks good

examples/01-basic/app.ts
donmccurdy: For now I think I'll just remove this from the examples, and include it in unit tests instead – but agreed that setting initial view state would be a good use case!

I'm tempted to wait on that until we feel OK that the API is no longer 'experimental', as putting it in the examples might encourage adoption too soon.

PR #223, opened 2025-07-22
feat(fetch-map): support for raster layers and updateTriggers

zbigg:
Extend parseMap/fetchMap with new functionality

* parseMap/fetchMap: support for raster layers configured by builder
* parseMap/fetchMap: add updateTriggers
* parseMap/fetchMap: add support for variable stroke width (fix)
* parseMap/fetchMap: expose domain (natural) and scaleDomain  (passed to d3 scale) as separate properties
* parseMap: don't mutate input (fix)
* expose getLayerDescriptor to get props of single layer
* vector expression evaluator moved here as private API (as it's prerequisite for performant raster layer rendering)

### DIFF
--- a/test/fetch-map/vec-expr-evaluator.test.ts
+++ b/test/fetch-map/vec-expr-evaluator.test.ts
@@ -1,0 +1,316 @@
+import {describe, it, expect} from 'vitest';
+import {
+  _createVecExprEvaluator as createVecExprEvaluator,
+  _ErrorCode as ErrorCode,
+  _validateVecExprSyntax as validateVecExprSyntax,
+  _VecExprResult as VecExprResult,
+} from '@carto/api-client';
+
+const repeat = (n, fn) => {
+  for (let i = 0; i < n; i++) {
+    fn();
+  }
+};
+
+describe('vecExprEvaluator', () => {
+  const band_1 = [1, 2, 3, 4, 5];
+  const band_2 = [10, 10, 10, 10, 10];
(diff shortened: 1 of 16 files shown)


test/fetch-map/vec-expr-evaluator.test.ts
```diff
@@ -0,0 +1,316 @@
...
+      });
+    });
+  });
+
+  // const PERFTEST_REPETITIONS = 4; // repeats, by default decreased not to slow down the tests
+  const PERFTEST_REPETITIONS = 15; // repeats, with 15-20 (at least on M1) you should see actual numbers reported by vitest
+  const PERFTEST_ARRAYS_SIZE = 200000; // size of arrays
+
+  // this test shows that it was necessary to vectorize the expression evaluation
+  describe(`performance ${PERFTEST_REPETITIONS} x evaluate expr on array ${PERFTEST_ARRAYS_SIZE}`, () => {
```
donmccurdy, line 202: Optional - we do have a separate benchmark suite in this repo...

https://github.com/CartoDB/carto-api-client/blob/main/scripts/benchmark.js

... so it could be a good option to move the benchmarks there, at some point. It's currently just one file but in other projects I've split each benchmark task into a separate file:

https://github.com/donmccurdy/glTF-Transform/tree/v4.2.0/benchmarks

donmccurdy (approved):

donmccurdy (approved):

PR #228, opened 2025-08-21
Trajectory source timestampRange

felixpalmer:
## Background

Follow-up to https://github.com/CartoDB/carto-api-client/pull/227

For trajectories, we need to know what the range of time values is in the dataset (much like `extent` in tilejson gives us the spatial extent). By using the WidgetSource we can obtain this from the API and make it easily available to the application developer. While in theory we could just return a function, e.g. `getTimestampRange` instead, this usage is expected to be so common it is best to encapsulate the `await` and save the user from having to deal with another async call

## Changes
- Add `timestampRange` to the `TrajectoryXXXSourceResponse`
- Tests


src/sources/types.ts
```diff
@@ -434,3 +434,8 @@ export type QueryResult = {
   rows: Record<string, any>[];
   schema: {name: string; type: string}[];
 };
+
+export type TimestampRangeResult = {
+  min: number | string;
+  max: number | string;
+} | null;
```
donmccurdy: Can we make these types a bit more strict? While getRange does support range and numeric columns I think, is that true of a trajectory timestamp column? If the result is null does that indicate an error in the source, which we could throw? If we can safely make it a little tighter here, user apps will have less they need to check/handle.

donmccurdy (approved):

src/sources/types.ts
felixpalmer: I am testing with two dataset, one with a numeric column and the other with a timestamp (which comes through as a string). So `getRange` certainly works in this case. The `null` comes from the `RangeResponse` type - I'm not sure why it is there.

All this considered, I wonder if we should just update:

```
 RangeResponse = {min: number; max: number} | {min: string; max: string}
```

RuBiCK: @gemini-code-assist review

felixpalmer: @gemini-code-assist I meant to update the `RangeResponse` type in `src/widget-sources/types.ts` and then just use that type.

donmccurdy: @felixpalmer `null` is a valid response from getRange(), likely if a table has no values. I guess it's a question of the desired behavior for trajectoryTableSource on an empty table. AFAIK other sources should gracefully "show nothing", so my earlier idea of throwing an error probably isn't right unless we know trajectory sources would have larger issues. 

I like the idea of updating both types to:

```ts
{
  min: number;
  max: number;
} | {
  min: string;
  max: string;
} | null;
```

felixpalmer: For a tile with no geometry data, an error isn't thrown, so I think it is valid to return `null`. If this corresponds to an empty table, it is useful to clearly inform the app

PR #229, opened 2025-08-25
feat(widgets): Add getAggregations() method

pcaro:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/505098
- Autolink: [sc-505098]

Create widgets out of pivoted tables where data is pre-aggregated in different numeric columns

Related: https://github.com/CartoDB/cloud-native/pull/21091

## Type of change

- [ ] Fix
- [X ] Feature
- [ ] Refactor
- [ ] Tests
- [ ] Documentation
- [ ] Security
- [ ] Chore

### DIFF
--- a/src/widget-sources/types.ts
+++ b/src/widget-sources/types.ts
@@ -278,7 +278,9 @@ export type TimeSeriesResponse = {
 export type HistogramResponse = number[];
 
 /** Response from {@link WidgetRemoteSource#getAggregations}. */
-export type AggregationsResponse = Record<string, number>;
+export type AggregationsResponse = {
+  rows: Record<string, number>[];
+};
 
 /** @experimental */
 export type ExtentResponse = {bbox: BBox};
--- a/src/widget-sources/types.ts
+++ b/src/widget-sources/types.ts
@@ -207,6 +207,22 @@ export interface TimeSeriesRequestOptions extends BaseRequestOptions {
   splitByCategoryValues?: string[];
 }
 
+/**
+ * Examples:
+ *   * aggregations with array syntax
+ *      * aggregations: [{column: 'pop_high', operation: 'sum', alias: 'high_pop'}, {column: 'pop_low', operation: 'avg'}]
+ *   * aggregations with string syntax
+ *      * aggregations: 'sum(pop_high) as high_pop, avg(pop_low) as avg_low'
+ *
+ * Options for {@link WidgetRemoteSource#getAggregations}.
+ */
+export interface AggregationsRequestOptions extends BaseRequestOptions {
+  /** Aggregations to compute. Can be an array of objects or a SQL string expression. */
+  aggregations: 
+    | { column: string; operation: AggregationType; alias?: string }[]
+    | string;
+}
+
 /** @experimental */
 export type ExtentRequestOptions = BaseRequestOptions;
 
@@ -260,6 +276,9 @@ export type TimeSeriesResponse = {
 
 /** Response from {@link WidgetRemoteSource#getHistogram}. */
 export type HistogramResponse = number[];
+
+/** Response from {@link WidgetRemoteSource#getAggregations}. */
+export type AggregationsResponse = Record<string, number>;
 
 /** @experimental */
 export type ExtentResponse = {bbox: BBox};
--- a/src/widget-sources/types.ts
+++ b/src/widget-sources/types.ts
@@ -218,8 +218,12 @@ export interface TimeSeriesRequestOptions extends BaseRequestOptions {
  */
 export interface AggregationsRequestOptions extends BaseRequestOptions {
   /** Aggregations to compute. Can be an array of objects or a SQL string expression. */
-  aggregations: 
-    | { column: string; operation: Exclude<AggregationType, 'custom'>; alias?: string }[]
+  aggregations:
+    | {
+        column: string;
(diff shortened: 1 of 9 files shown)


src/widget-sources/types.ts
```diff
@@ -261,5 +281,10 @@ export type TimeSeriesResponse = {
 /** Response from {@link WidgetRemoteSource#getHistogram}. */
 export type HistogramResponse = number[];
 
+/** Response from {@link WidgetRemoteSource#getAggregations}. */
+export type AggregationsResponse = {
+  rows: Record<string, number>[];
```
donmccurdy: Is it correct that the value of the row will always be a number? I'm imagining min/max could be used for string and timestamp columns, for example?

src/widget-sources/widget-tileset-source-impl.ts
```diff
@@ -391,6 +393,78 @@ export class WidgetTilesetSourceImpl extends WidgetSource<WidgetTilesetSourcePro
...
+        result[aggregationKey] = aggregationFunctions.max(
+          filteredFeatures,
+          column
+        );
+      } else {
+        throw new Error(
+          `Unsupported aggregation operation: ${String(operation)}`
+        );
+      }
+    }
```
donmccurdy: Could this be condensed to something like:

```suggestion
      const targetOperation = aggregationFunctions[operation];
      result[aggregationKey] = targetOperation(filteredFeatures, column)
```

I think type safety is enough without a runtime check in this case, see `getFormula`, but I wouldn't object to one-line assert if you prefer. :)

src/widget-sources/widget-tileset-source-impl.ts
```diff
@@ -391,6 +393,78 @@ export class WidgetTilesetSourceImpl extends WidgetSource<WidgetTilesetSourcePro
...
+      if ((column && column !== '*') || operation !== AggregationTypes.Count) {
+        assertColumn(this._features, column);
+      }
+
+      const aggregationKey = alias || `${operation}_${column}`;
+      const aliasKey = aggregationKey.toLowerCase();
+      if (usedAliases.has(aliasKey)) {
+        throw new Error(`Duplicate aggregation alias: ${aggregationKey}`);
+      }
+      usedAliases.add(aliasKey);
```
donmccurdy: If column is an empty string or '*' this might give unexpected results, I think. Do we know what alias the Model API chooses in those cases, and could we do the same locally?

src/widget-sources/types.ts
pcaro: Talked with @jacrisol  we are going to allow number, string and null (this last one in case filter provides no columns for instance)

src/widget-sources/widget-tileset-source-impl.ts
pcaro: Done

src/widget-sources/types.ts
pcaro: Done

src/widget-sources/widget-tileset-source-impl.ts
pcaro: Alias is mandatory now. Done.

donmccurdy: I don't see this change pushed yet, the alias type is optional and has a fallback to `${operation}_${column}` above — possibly forgot to push?

donmccurdy (approved):
One issue remaining, probably the alias field should be required — otherwise LGTM!

src/widget-sources/widget-tileset-source-impl.ts
pcaro: Done.

test/widget-sources/widget-remote-source.test.ts
```diff
@@ -41,6 +42,222 @@ test('constructor', () => {
   });
 });
 
+test('getAggregations - normaliza claves y mezcla alias/por defecto', async () => {
+  const widgetSource = new WidgetTestSource({
+    accessToken: '<token>',
+    connectionName: 'carto_dw',
+  });
+
+  // Simula un DW que devuelve claves con mayúsculas variadas y/o sin alias
```
zbigg: please translate these comments to english

src/widget-sources/widget-remote-source.ts
```diff
@@ -394,6 +396,36 @@ export abstract class WidgetRemoteSource<
...
+    const {
+      signal,
+      filters = this.props.filters,
+      filterOwner,
+      spatialFilter,
+      spatialFiltersMode,
+      aggregations,
+    } = options;
+
+    type AggregationsModelResponse = {rows: Record<string, number>[]};
```
zbigg: this type seems wrong - no string aggregation and redundant, as there is `AggregationsResponse`

src/widget-sources/widget-tileset-source-impl.ts
pcaro: Changed test to reflect alias are now mandatory

test/widget-sources/widget-remote-source.test.ts
pcaro: Done

src/widget-sources/widget-remote-source.ts
pcaro: Changed. Thx

PR #241, opened 2025-09-19
fix(fetchMap): support quantile colorMap with two colors [sc-511580]

zbigg:
fetchMap - support quantile color scale with _two_ colors, deriving median from quantiles[4]

Extra, don't crash when there are no quantiles, just return default domain - `[min,max]`

https://app.shortcut.com/cartoteam/story/511580

### DIFF
--- a/test/fetch-map/layer-map.spec.ts
+++ b/test/fetch-map/layer-map.spec.ts
@@ -5,6 +5,12 @@ import {
   getLayerProps,
   _domainFromValues,
 } from '@carto/api-client';
+import {rgb} from 'd3-color';
+
+const hexToRGBA = (c: any) => {
+  const {r, g, b} = rgb(c);
+  return [r, g, b, 255];
+};
 
 describe('layer-map', () => {
   const colors = [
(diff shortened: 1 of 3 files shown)


test/fetch-map/layer-map.spec.ts
```diff
@@ -5,6 +5,12 @@ import {
   getLayerProps,
   _domainFromValues,
 } from '@carto/api-client';
+import {rgb} from 'd3-color';
```
donmccurdy, line 8: I think `d3-color` is only in the repository as an indirect dependency now, perhaps let's add it to the devDependencies? Or just in-lining a 1-3 line function here could be fine too.

donmccurdy (approved):

test/fetch-map/layer-map.spec.ts
zbigg: It's still used in `layer-map.ts` that is main code

But i think it's good idea to get rid of it - we only support basic `#rrggbb` variant which is easy to parse.

zbigg: Another observation - `d3-array` is used by code only used by `domainFromValues` which is remnant from document mode (i believe)

zbigg: >is only in the repository as an indirect dependency

Ok, i misread. Good catch. I'll try to cleanup it in next PR.

PR #255, opened 2025-11-06
fix(fetchMap): Add tag-mapId to requests 

zbigg:
Add `tag-mapId` to requests initiated by fetchMap and it's internal functions like `_fillInMapDatasets` and `_fillInTileStats`

### DIFF
--- a/src/fetch-map/fetch-map.ts
+++ b/src/fetch-map/fetch-map.ts
@@ -34,9 +34,11 @@ async function _fetchMapDataset(
       accessToken: context.accessToken!,
       apiBaseUrl: context.apiBaseUrl,
       maxLengthURL: context.maxLengthURL,
-      tags: {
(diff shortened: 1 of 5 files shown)


src/fetch-map/fetch-map.ts
```diff
@@ -26,14 +26,19 @@ async function _fetchMapDataset(
   const cache: {value?: number} = {};
   const configuredSource = configureSource({
     dataset,
-    filters: isRemoteCalculationSupported(dataset) ? filters : undefined,
+    filters: !isEmptyObject(filters) && isRemoteCalculationSupported(dataset) ? filters : undefined,
```
zbigg: Builder already skips empty filters, so adding it here allows to reuse responses  between those created by builder and by fetchMap in one cache.

padawannn (approved):

PR #259, opened 2025-12-03
chore(dependabot): ignore patch updates

felixpalmer:
To reduce the dependabot notices, ignore patch releases


padawannn:
Closed because the discussion comes from 2025. If we want to push it re-open the discussion

PR #273, opened 2026-03-24
fix(widgetSources): add offset request option and total for pagination

Zulcom:
Pick API returns the total entity count only when both offset and limit are provided. Add support for the offset option.


zbigg (approved):
lgtm, but please add changelog entry (as in other carto-api-client) PRs

PR #281, opened 2026-04-23
feat(parseMap): add compile custom aggregation

Zulcom:
Adds the api-client side of the custom aggregation feature for spatial-index and geometry-aggregated layers. The Builder (workspace-www) now stores `{channel}AggregationExp` on saved maps. `parseMap` resolves those expressions to the same content-addressed tile column alias that Builder used when generating the `aggregationExp` SQL. That way, deck reads the right property and the rendering scale follows the user-defined min and max.

## What's included

* `compileCustomAggregation(expression, { provider })`
  New public helper. Uses FNV-1a 32-bit hashing to produce an 8-character hex string in the format `custom_agg_<hash>`. The result is uppercased for Snowflake. Input is normalised for whitespace before hashing. Throws if the expression is empty or only whitespace.

* Type updates
  `VisualChannelField.accessorKey?: string`
  `VisConfig.{channel}AggregationExp?: string`
  `VisConfig.{channel}AggregationDomain?: [number, number]`
  These apply to all six channels: color, strokeColor, size, height, radius, and weight.

* `resolveCustomAggregation()` in `parse-map.ts`
  When `{channel}Aggregation === 'custom'`, it recomputes the alias using `compileCustomAggregation` and assigns it to `accessorKey` on the channel field. The domain is passed through as `domainOverride`.

* Accessors
  `getColorAccessor` and `getSizeAccessor` now use `accessorKey` instead of `colorColumn` or `name` when present.
  `calculateLayerScale` uses `domainOverride` instead of tilestats when available.
  `colorRange.colorMap` still takes priority over both, so explicit break values always win.

* `'lineWidth'` added to the `ScaleKey` union - Fixes an existing type gap.

## Tests

17 tests in `test/fetch-map/parse-map.spec.ts`, plus helper tests in `test/fetch-map/aggregation.test.ts` and accessor override tests in `test/fetch-map/layer-map.spec.ts`.

Coverage includes:

* Helper behavior: provider casing across all supported providers, whitespace normalisation (spaces and tabs), error on empty input, deterministic output, and uniqueness.
* `parseMap` for each channel: color, strokeColor, size to lineWidth, height with enable3d, radius, and weight. Each test checks that `accessorKey` is set and that the domain override is applied.
* Standard aggregation remains unchanged. `accessorKey` stays undefined and existing accessor logic is used.
* Mixed case with custom color and standard size on the same layer. Only the custom channel gets an `accessorKey`.
* Precedence rules: `colorRange.colorMap` break values override `colorAggregationDomain`.
* Defensive checks: empty or whitespace-only expressions with `colorAggregation === 'custom'` do not set `accessorKey`.
* Coexistence of `colorColumn` and `accessorKey`. The accessor uses the custom alias while the scale keeps the configured `scaleType`.

## Impact

* No runtime changes unless the feature is used. `accessorKey` is only set when a layer has `{channel}Aggregation === 'custom'` with a valid expression.
* Builder uses this directly. External `@carto/api-client` users get additional optional fields in the type definitions.

### DIFF
--- a/src/fetch-map/layer-map.ts
+++ b/src/fetch-map/layer-map.ts
@@ -342,9 +342,17 @@ function findAccessorKey(keys: string[], properties: any): string[] {
 }
 
 export function getColorAccessor(
-  {name, colorColumn}: VisualChannelField,
+  {name, colorColumn, accessorKey}: VisualChannelField,
   scaleType: ScaleType,
-  {aggregation, range}: {aggregation?: string; range: ColorRange},
+  {
+    aggregation,
+    range,
+    domainOverride,
+  }: {
+    aggregation?: string;
+    range: ColorRange;
+    domainOverride?: [number, number];
+  },
   opacity: number | undefined,
   data: TilejsonResult
 ): {
@@ -353,15 +361,23 @@ export function getColorAccessor(
   scaleDomain: number[] | string[];
   range: string[];
 } {
+  // accessorKey (custom-agg alias) wins over colorColumn (legacy identity-color
+  // column), which wins over name (standard field).
+  const effectiveName = accessorKey ?? colorColumn ?? name;
+  const effectiveScaleType =
+    colorColumn && !accessorKey ? 'identity' : scaleType;
   const {scale, domain} = calculateLayerScale(
-    colorColumn || name,
-    colorColumn ? 'identity' : scaleType,
+    effectiveName,
+    effectiveScaleType,
     range,
-    data
+    data,
+    domainOverride
   );
   const alpha = opacityToAlpha(opacity);
 
-  let accessorKeys = getAccessorKeys(colorColumn || name, aggregation);
+  let accessorKeys = accessorKey
+    ? [accessorKey]
+    : getAccessorKeys(colorColumn || name, aggregation);
   const accessor = (properties: any) => {
     if (!(accessorKeys[0] in properties)) {
       accessorKeys = findAccessorKey(accessorKeys, properties);
@@ -383,13 +399,20 @@ export function calculateLayerScale(
   name: string,
   scaleType: ScaleType,
   range: ColorRange,
-  data: TilejsonResult
+  data: TilejsonResult,
+  domainOverride?: [number, number]
 ): {scale: D3Scale; domain: string[] | number[]} {
   let scaleDomain: number[] | string[] | undefined;
   let scaleColors: string[] = [];
   const {colors} = range;
 
-  const domain = calculateDomain(data, name, scaleType, colors.length);
+  // colorMap (explicit break values from Custom classification) wins over
+  // domainOverride (user Min/Max for Custom aggregation). colorMap defines
+  // its own domain, so the override does not apply.
+  const domain =
+    domainOverride && !range.colorMap
+      ? domainOverride
+      : calculateDomain(data, name, scaleType, colors.length);
   if (scaleType !== 'identity') {
     if (range.colorMap) {
       const {colorMap} = range;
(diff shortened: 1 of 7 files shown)


Zulcom:
> I have no feedback to provide.

AI imitates humans :D

src/fetch-map/layer-map.ts
```diff
@@ -597,14 +626,42 @@ export function getDefaultAggregationExpColumnAliasForLayerType(
...
   name: string,
   aggregation: string,
   provider: ProviderType
 ) {
-  const columnAlias = `${name}_${aggregation}`;
-  return provider === 'snowflake' ? columnAlias.toUpperCase() : columnAlias;
+  return applyProviderCase(`${name}_${aggregation}`, provider);
+}
+
+export function compileCustomAggregation(
```
jmgaya, line 655: Hmm, does it mean we're going to keep "current expressions" in Builder as they are right now? I mean, this solves the problem for "free-text ones", but what happens for those you select the **operation** and then the **column**?

Zulcom: I think this would benefit deduping, but can harm debugging and anyway deserves separate story and qa after dealing with custom ones

src/fetch-map/layer-map.ts
```diff
@@ -597,14 +626,42 @@ export function getDefaultAggregationExpColumnAliasForLayerType(
...
+  exp: string,
+  opts: {provider: ProviderType}
+): string {
+  const normalized = normalizeExpression(exp);
+  if (!normalized) {
+    throw new Error('Custom aggregation expression must not be empty');
+  }
+  const hash = hashString(normalized);
+  return applyProviderCase('custom_agg_' + hash, opts.provider);
```
zbigg, line 664: just wondering -  maybe we should prefix all "Carto-derived" symbols with `_carto` ?

(i know it makes sql even bigger, but we'll be sure that we don't have any conflicts and it's clear where does this symbols comes from)

src/fetch-map/parse-map.ts
```diff
@@ -250,6 +253,37 @@ function createStyleProps(config: MapLayerConfig, mapping: any) {
...
+  expression: string | undefined;
+  domain: [number, number] | undefined;
+  providerId: ProviderType;
+}): {
+  field: VisualChannelField | undefined;
+  aggregation: string | undefined;
+  domainOverride: [number, number] | undefined;
+} {
+  if (aggregation !== 'custom') {
+    return {field, aggregation, domainOverride: undefined};
```
zbigg, line 274: Idea: Can we  resolve `accessorKey` for normal aggretation here and maybe, possibly remove this dirty (imo) logic relying on `getAccessorKeys`?

src/fetch-map/layer-map.ts
Zulcom: A user column called `custom_agg_3a7f2b1c` colliding by accident is essentially zero, and the same conflict surface exists today for the standard `${name}_${aggregation}` aliases (e.g., `population_avg`), which also don't have a `_carto` prefix. So if we want a CARTO-namespaced convention, it should be a broader cleanup that covers both standard and custom paths together, behind a Builder + api-client + saved-map migration. 

Happy to create a follow-up ticket to do that, keeping it out of this PR so we don't widen scope or break existing maps

src/fetch-map/parse-map.ts
Zulcom: Actually, we can pre-resolve accessorKey for normal aggregation here, too, and delete `getAccessorKeys` and `findAccessorKey`. But getColorAccessor/getSizeAccessor are on the hot path for **every layer in every saved map**, and removing the runtime fallback means any edge case we missed (a layer type, a provider quirk, a user-aggregated source where aggregation happens to be set for widgets already) would silently break rendering with no graceful recovery. In my understanding, this is a feature-adding initiative, not a breaking-everything refactor with several risks

zbigg: I'm quite confident* that there are no edge cases.

It's baseically

* normal aggregation -> `{name}_{agg}` - possibly uppercase
* custom aggregstion -> new one, hashed
* normal field -> `name`

The only thing* (as usual) - that i'm not surely fully understanding is forced upper case. Is this only snowlake and is it always, etc.

(this is actually very complex, because snowflake has support for low/fixed case columns and in that case it doesn't force them uppercase, but we don't support it very well)

jmgaya (approved):

PR #298, opened 2026-06-12
feat(auth): session auth mode for same-origin proxy deployments [sc-556407]

jatorre:
## Why

Some deployments must never expose a CARTO access token to JavaScript: the credential lives server-side and requests are authenticated by a session cookie behind a same-origin proxy. The first consumer is **CARTO hosted apps**, where every API call goes through `/app/_proxy/<slug>/v3/...` and the platform attaches the viewer's credential server-side. The same shape serves selfhosted single-origin deployments and embedded contexts.

Today the client requires an `accessToken` and attaches `Authorization: Bearer …` unconditionally, which defeats that model (a session proxy must see the header *absent* to engage).

## What

Additive `authMode?: 'token' | 'session'` on the shared auth options (sources, `query`, widget sources, models). Default `'token'` preserves today's behavior exactly. With `authMode: 'session'`:

- **No Authorization header** is sent on any request. All header construction now goes through a single `buildAuthHeaders()` helper (was 6 inline sites), which also validates the options: token mode without a token, or session mode *with* a token, throws a clear error instead of sending `Bearer undefined`.
- **`credentials: 'same-origin'`** is passed to `fetch()` (new `credentials` passthrough in `requestWithParameters`).
- **Tilejson post-processing rewrites URLs onto `apiBaseUrl`**: the map-instantiation `dataUrl` and the `tiles[]` templates point at the tenant API host, which the browser can't reach directly in this mode. `TilejsonResult.accessToken` is left unset (type now optional) so consumers don't fabricate a bearer; `createLoadOptions` (fetchMap/parse-map) now omits the Authorization header when no token is present instead of emitting a literal `Bearer undefined`.
- **Relative `apiBaseUrl` support** in `requestWithParameters` URL helpers (`new URL()` previously threw on `/app/_proxy/...` bases).

New exports: `buildAuthHeaders`, `getAuthCredentials`, `rewriteUrlForSessionMode`, `AuthMode`, `AuthOptions`.

Out of scope (follow-up): full session-mode support inside `fetchMap` for Builder maps — it already tolerates a missing token; deep rewriting of dataset tile URLs there is a larger surface.

## Usage

```ts
const data = await vectorTableSource({
  connectionName: 'carto_dw',
  tableName: 'my_table',
  apiBaseUrl: '/app/_proxy/my-app', // same-origin proxy
  authMode: 'session',              // no token in JS
});
```

## Testing

- New `test/api/auth.test.ts` covering the helpers and validation errors.
- `base-source.test.ts`: session mode asserts no Authorization header, `same-origin` credentials, instantiation + dataUrl routed through the proxy base, `tiles[]` rewritten, no `accessToken` on the result; plus a token-mode regression test.
- Full suite: 56 files / 449 tests passing, typecheck and lint clean.

## Notes for review

- `TilejsonResult.accessToken` changed `string` → `string | undefined` (set only in token mode, as before). Strict downstream consumers may need a guard.
- `buildAuthHeaders` introduces a client-side error for a missing token (previously the request went out as `Bearer undefined` and failed server-side with a 401) — strictly better DX, but it is a behavior change for misconfigured callers.
- deck.gl's CARTO tile layers currently build `Authorization: Bearer ${tileJSON.accessToken}` unconditionally; with this PR's unset token they'd send `Bearer undefined`. The hosted-apps proxy tolerates that server-side, and a one-line conditional fix to @deck.gl/carto is planned as a companion.

[sc-556407]

### DIFF
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -1,7 +1,3 @@
-// deck.gl
-// SPDX-License-Identifier: MIT
-// Copyright (c) vis.gl contributors
-
 /**
  * How requests to the CARTO APIs are authenticated.
  *
@@ -77,8 +73,20 @@ export function rewriteUrlForSessionMode(
   url: string,
   apiBaseUrl: string
 ): string {
+  let origin: string;
+  try {
+    const parsed = new URL(url);
+    if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+      return url;
+    }
+    origin = parsed.origin;
+  } catch {
+    // Relative URL — already same-origin, nothing to rewrite.
+    return url;
+  }
+  // Keep the path+query as a raw slice rather than re-serializing through
+  // URL: tile URL templates contain literal placeholders ({z}/{x}/{y}) that
+  // URL#pathname would percent-encode, breaking template substitution.
   const base = apiBaseUrl.replace(/\/+$/, '');
-  // Use a replacement function so any `$`-patterns in apiBaseUrl (`$&`, `$1`,
-  // …) are treated literally rather than interpreted by String.replace.
-  return url.replace(/^https?:\/\/[^/?#]+/, () => base);
+  return base + url.slice(origin.length);
 }
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -77,5 +77,8 @@ export function rewriteUrlForSessionMode(
   url: string,
   apiBaseUrl: string
 ): string {
-  return url.replace(/^https?:\/\/[^/?#]+/, apiBaseUrl.replace(/\/+$/, ''));
+  const base = apiBaseUrl.replace(/\/+$/, '');
+  // Use a replacement function so any `$`-patterns in apiBaseUrl (`$&`, `$1`,
+  // …) are treated literally rather than interpreted by String.replace.
+  return url.replace(/^https?:\/\/[^/?#]+/, () => base);
 }
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -1,0 +1,81 @@
+// deck.gl
+// SPDX-License-Identifier: MIT
+// Copyright (c) vis.gl contributors
+
+/**
+ * How requests to the CARTO APIs are authenticated.
+ *
+ * - `'token'` (default): every request carries `Authorization: Bearer <accessToken>`.
+ * - `'session'`: requests carry NO Authorization header and are sent with
+ *   `credentials: 'same-origin'`. Authentication is delegated to the server
+ *   behind `apiBaseUrl` — typically a same-origin proxy that attaches the
+ *   credential server-side from a session cookie. Used by CARTO hosted apps
+ *   (`apiBaseUrl: '/app/_proxy/<slug>'`) and other single-origin deployments
+ *   where the access token must never be exposed to JavaScript.
+ */
+export type AuthMode = 'token' | 'session';
+
+/**
+ * Authentication options shared by sources, queries, widget sources and
+ * fetchMap:
+ *
+ * - token mode (default): `accessToken` is required.
+ * - session mode: `authMode: 'session'` and no `accessToken`.
+ *
+ * Validated at request time by {@link buildAuthHeaders}.
+ */
+export type AuthOptions = {
+  /** Carto platform access token. Required unless `authMode` is `'session'`. */
+  accessToken?: string;
+  /** @default 'token' */
+  authMode?: AuthMode;
+};
+
+/**
+ * Builds the Authorization headers for a request: a Bearer header in token
+ * mode, no auth headers at all in session mode (the server-side session
+ * middleware must see the header absent to engage). Also validates the
+ * options, so every request path fails fast on a misconfiguration.
+ */
+export function buildAuthHeaders(options: AuthOptions): Record<string, string> {
+  if (options.authMode === 'session') {
+    if (options.accessToken) {
+      throw new Error(
+        `accessToken must not be provided with authMode: 'session' — ` +
+          `authentication is delegated to the server behind apiBaseUrl`
+      );
+    }
+    return {};
+  }
+  if (!options.accessToken) {
+    throw new Error(
+      `accessToken is required (or pass authMode: 'session' to authenticate ` +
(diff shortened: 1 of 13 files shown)


jatorre:
## Review follow-up: all three Gemini comments addressed

I reviewed the Gemini Code Assist feedback — all three points are valid. Below are the fixes (verified locally: full suite **450 passed / 2 skipped** incl. one new test, `tsc` typecheck clean, `eslint` clean). They're committed locally but I couldn't push from this session (the environment only has fork-read access to this repo), so posting here for you to apply on `feat/session-auth-mode`.

### 1. (High) `src/models/model.ts` — assertion throws in session mode
`assert(accessToken, …)` runs before the request is built and throws when `authMode: 'session'` (no token). Rather than guard the assert, I **removed it**: auth is already validated centrally by `buildAuthHeaders` (via `makeCall`), with a clearer message, exactly as `base-source.ts` does. The now-unused `accessToken` is dropped from the destructure.

```diff
-  const {type, apiVersion, apiBaseUrl, accessToken, connectionName, clientId} =
-    source;
+  const {type, apiVersion, apiBaseUrl, connectionName, clientId} = source;

   assert(apiBaseUrl, 'executeModel: missing apiBaseUrl');
-  assert(accessToken, 'executeModel: missing accessToken');
+  // Auth (token vs session mode) is validated centrally by buildAuthHeaders,
+  // invoked from makeCall below — no separate accessToken assertion here, so
+  // that session mode (no accessToken) is not rejected before the request.
   assert(apiVersion === V3, 'executeModel: SQL Model API requires CARTO 3+');
```

### 2. (Medium) `src/api/auth.ts` — `$`-patterns in `apiBaseUrl`
`String.replace` with a string replacement interprets `$&`, `$1`, … Use a replacement function so the base is inserted literally.

```diff
-  return url.replace(/^https?:\/\/[^/?#]+/, apiBaseUrl.replace(/\/+$/, ''));
+  const base = apiBaseUrl.replace(/\/+$/, '');
+  // Use a replacement function so any `$`-patterns in apiBaseUrl (`$&`, `$1`,
+  // …) are treated literally rather than interpreted by String.replace.
+  return url.replace(/^https?:\/\/[^/?#]+/, () => base);
```
Plus a regression test in `test/api/auth.test.ts`:
```ts
test('treats $-patterns in the base literally', () => {
  expect(
    rewriteUrlForSessionMode('https://host.com/v3/sql', '/proxy/$&/$1')
  ).toBe('/proxy/$&/$1/v3/sql');
});
```

### 3. (Medium) `src/api/request-with-parameters.ts` — protocol-relative URLs
`//host/...` carries a host and must not be treated as relative (it would lose the host against `RELATIVE_ORIGIN`).

```diff
-  const isRelative = baseUrlString.startsWith('/');
+  // Protocol-relative URLs ('//host/...') carry a host and must not be treated
+  // as relative, or the host would be lost when resolved against RELATIVE_ORIGIN.
+  const isRelative =
+    baseUrlString.startsWith('/') && !baseUrlString.startsWith('//');
```

---
_Generated by [Claude Code](https://claude.ai/code)_

jatorre:
All three fixes from the comment above are now applied and pushed (`72c9f24`): executeModel no longer asserts accessToken pre-request (auth validated centrally in buildAuthHeaders), rewriteUrlForSessionMode inserts the base via a replacement function (with the $-pattern regression test), and protocol-relative URLs are no longer mis-detected as relative. Verified: 450 passed / 2 skipped, typecheck, lint, and prettier clean.

jatorre:
@zbigg @felixpalmer this is ready for review — adds an additive `authMode: 'session'` so apps behind a same-origin proxy (first consumer: CARTO hosted apps) can use api-client without exposing a token to JS. Gemini's three review points are already addressed (`72c9f24`); CI green on Node 18/20/22. I don't have permission to formally request reviewers on this repo, hence the ping — happy to walk you through the hosted-apps context if useful.

src/api/auth.ts
```diff
@@ -0,0 +1,84 @@
+// deck.gl
```
zbigg: nit: not needed, this code is not derived/copied from deck.gl afair
(that's why some existing files have this header - they were copied from @deck.gl/carto)

src/api/auth.ts
```diff
@@ -0,0 +1,84 @@
...
+ * the map-instantiation `dataUrl`, which always point at the tenant API host)
+ * onto the configured `apiBaseUrl`. In session mode the browser cannot reach
+ * the API host directly — every request must go through the same-origin
+ * proxy that `apiBaseUrl` points at. Relative URLs are returned unchanged.
+ */
+export function rewriteUrlForSessionMode(
+  url: string,
+  apiBaseUrl: string
+): string {
+  const base = apiBaseUrl.replace(/\/+$/, '');
```
zbigg, line 90: it would be easier to understand if we just parse using `URL`.

rationale: using regexp  for parsing URL is tricky and we have "new URL(...)" already in javascript library

i mean something like this (pseudo code, not checked)
```
const x = new URL(url)

return `${base}/${x.path}${x.search}`
```

zbigg (approved):
lgtm, few comments only

jatorre:
Both nits addressed in `4e2dbb6`:

1. **deck.gl header dropped** from the two new files (`src/api/auth.ts`, `test/api/auth.test.ts`) — they're original, not copied from @deck.gl/carto.
2. **`rewriteUrlForSessionMode` now parses with `new URL()`** instead of the regex. One twist worth flagging: the naive `base + parsed.pathname + parsed.search` re-serialization percent-encodes the literal `{z}/{x}/{y}` placeholders in tile URL templates (`/%7Bz%7D/...`), which would break deck.gl's template substitution — so the implementation uses `URL` for parsing/validation but keeps the path+query as a raw slice: `base + url.slice(parsed.origin.length)`. Comment in code explains it.

Full suite still green (450 passed, typecheck/lint/prettier clean).

zbigg (approved):

jatorre:
@zbigg thanks for the review! Nits are addressed (`4e2dbb6`) and everything is green — I don't have write access on this repo, so whenever you're happy could you merge it? If a release is cheap to cut afterwards, that would also unblock the consumer we built this for (CARTO hosted apps' PS-app ports pin against the published package).

PR #300, opened 2026-06-13
feat(sources): use inlined tilejson document when available

jatorre:
> 📊 Part of **Modernizing Maps Performance** — [epic sc-556570](https://app.shortcut.com/cartoteam/epic/556570) · [team deck](https://docs.google.com/presentation/d/18pQsKAE6YrHNexIUAqltV4tmnFOsy5YabgZ5G2YmaaU/edit)

## What

Map sources currently make two requests to render: **instantiate**, then **GET the tilejson** document the instantiation response points at (`tilejson.url`). This adds the client half of an opt-in optimization: when the server embeds the tilejson document in the instantiation response, use it directly and skip the follow-up request — **one round trip less per source on every map load**.

`baseSource` now:
1. Sends `inlineTilejson=true` on every instantiation request.
2. When the response carries `tilejson.data`, uses it directly instead of fetching `tilejson.url`.

## Why it's safe to merge ahead of the server

The change is fully backward-compatible in both directions:

- **Old server / non-inlining endpoint:** the unknown `inlineTilejson` query param is ignored and `tilejson.data` is absent, so the existing fetch-the-URL path runs exactly as today.
- **New server:** `tilejson.data` is present and the second request is skipped.

So this can ship before, after, or alongside the server change with no coordination required.

## Notes

- Session-mode proxy rewriting of the tile templates applies to the inlined document the same as the fetched one (the `sessionMode` block runs on `json` regardless of branch).
- The `cache.value` extraction still reads `?cache=` from `tilejson.url[0]` (always present); the server emits the same cache hash inside the inlined tile templates, so tiles fetched via the inlined path stay cache-consistent with the URL path.
- `inlineTilejson` is on by default here (the client always needs the tilejson document, so inlining is strictly a win). Easy to gate behind an option if you'd prefer it opt-in — let me know.

## Server side

maps-api embeds the document when called with `?inlineTilejson=true`: **CartoDB/cloud-native#25462** (sc-556572). Verified there that the inlined document is byte-identical to what a client fetches from `tilejson.url`, including the tile cache hash.

## Test plan

- New tests in `vector-table-source.test.ts`: the instantiation request carries `inlineTilejson=true`; when the response includes `tilejson.data`, `fetch` is called only once (follow-up skipped) and the returned tilejson is the embedded document.
- Full suite green (56 files, 452 tests, no type errors), lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- feat(sources): opt into `inlineTilejson` on map instantiation and use the embedded tilejson document when present, skipping the follow-up tilejson request — one round trip less per source. Falls back to fetching the URL on servers that don't support it (maps-api sc-556572). (#300)
 - feat(widgetSources): support `featureIds` / `geometryType` request options to filter widget aggregations by a feature selection without relying on the synthetic `_carto_feature_id` column (#294)
 
 ### 0.5.29
--- a/src/sources/base-source.ts
+++ b/src/sources/base-source.ts
@@ -54,7 +54,16 @@ export async function baseSource<UrlParameters extends Record<string, unknown>>(
     ...options.headers,
   };
   const credentials = getAuthCredentials(options.authMode);
-  const parameters = {client: clientId, ...options.tags, ...urlParameters};
+  // Opt into the server embedding the tilejson document in the instantiation
+  // response (maps-api sc-556572) so the follow-up GET below can be skipped.
+  // Safe against servers that don't support it: the unknown query param is
+  // ignored and `tilejson.data` is simply absent, falling back to the fetch.
+  const parameters = {
+    client: clientId,
+    inlineTilejson: true,
+    ...options.tags,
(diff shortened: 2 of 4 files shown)


jatorre:
Addressed the gemini review in `76189ec`:

- **🔴 in-place mutation of the cached embedded document** → `json = tilejson.data` then mutating `json` (tile URLs in sessionMode, `accessToken`, `schema`) mutated the document embedded in the instantiation response, which is cacheable — a reused response would then operate on an already-rewritten object (double-rewritten tile URLs in sessionMode). Now shallow-copies (`json = {...tilejson.data}`) before the rewrites; the mutations are all top-level reassignments, so a shallow copy suffices.

Test: the embedded document is left unmutated by the source call. (Built as a standalone literal rather than spread from the shared `MOCK_TILESET_RESPONSE`, which the non-inline fetch path mutates in place — a separate pre-existing test-hygiene quirk worth a follow-up.) Full suite 453/453; build + prettier + eslint clean.

PR #301, opened 2026-06-13
feat(sources): client-side tiler for the small-source fast path [sc-556576]

jatorre:
> 📊 Part of **Modernizing Maps Performance** — [epic sc-556570](https://app.shortcut.com/cartoteam/epic/556570) · [team deck](https://docs.google.com/presentation/d/18pQsKAE6YrHNexIUAqltV4tmnFOsy5YabgZ5G2YmaaU/edit)

## Summary

Client half of the maps-api **small-source fast path** (server story sc-556576 / cloud-native #25467). When a source is small the server collapses it to one full-resolution z0 tile (tilejson `maxzoom: 0`, tile URL `full=true`) instead of a pyramid — one warehouse query ever, CDN-shared, then zero warehouse work on pan/zoom. The client fetches that single tile once and answers deck.gl's per-tile requests **locally**, re-applying the server's per-zoom reduction (otherwise deck.gl's native overzoom draws every feature at every zoom and low-zoom density blows up — the failure the server story called out).

This PR adds the reusable slicer (pure functions, exported from `@carto/api-client`):

- `isFullSourceTilejson(tilejson)` — detect the `maxzoom: 0` + `full=true` signal.
- `buildFullTileIndex(features)` — index the full tile's features once (bbox + representative point + size); reused for every subsequent local slice.
- `sliceFullTile(index, {z,x,y}, {geomType, tileResolution})` — bbox-cull, then sub-pixel cull + biggest-first cap for lines/polygons, or a Web-Mercator grid aggregation for points (emitting the `_carto_point_density` count the tilers produce).

**Parity by construction:** the reduction formulas (`getToleranceForSimplify` / `getPointsAggregationLevel` / `getMaxFeaturesByResolution`, + the Mercator grid math) are ported verbatim from the maps-api dynamic tiler, so a locally-sliced tile matches what a per-tile request would have returned.

## Tests

8 unit tests (`test/sources/client-side-tiler.test.ts`): full-source detection, z0/0/0 bbox, low-zoom cluster → 1 feature + density count, deep-zoom → points stay separate, polygon bbox-cull, per-resolution cap. Full `test/sources` suite green (54).

## Not yet end-to-end — remaining under the same story

1. A `fullTiles` opt-in on the vector source so the client requests the full path (small carto-api-client change).
2. Wiring `sliceFullTile` into the deck.gl CARTO `VectorTileLayer` `getTileData` (a deck.gl PR) + visual A/B vs per-tile serving.

This PR is the tested, reusable core those two consume. Server counterpart: cloud-native #25467.

### DIFF
--- a/src/sources/client-side-tiler.ts
+++ b/src/sources/client-side-tiler.ts
@@ -20,6 +20,11 @@ export type SliceableGeomType = 'points' | 'lines' | 'polygons';
 
 const EARTH_RADIUS_METERS = 6378137;
 const EARTH_CIRCUMFERENCE_METERS = 2 * Math.PI * EARTH_RADIUS_METERS; // 40075016.6856
+// These three mirror maps-api *settings*, not constants: MAPS_API_V3_DYNAMIC_TILES_
+// POINTS_AGGREGATION_LEVEL, _MAX_POLYGON_FEATURES, _MAX_LINES_FEATURES. The values
+// below are the server defaults. If a deployment overrides those env vars the
+// server tiler and this client slicer drift apart (locally-sliced tiles stop
+// matching per-tile requests); keep them in sync if the server defaults change.
 const DYNAMIC_TILES_POINTS_AGGREGATION_LEVEL = 8;
 const DYNAMIC_TILES_MAX_POLYGON_FEATURES = 5000;
 const DYNAMIC_TILES_MAX_LINES_FEATURES = 10000;
@@ -57,11 +62,13 @@ export function tileToBBox(
   return [west, tileLat(y + 1), east, tileLat(y)];
 }
 
-/** Simplification tolerance for a zoom (server parity). Degrees by default. */
+/** Simplification tolerance for a zoom (server parity). Meters by default. */
 export function getToleranceForSimplify(
   tileResolution: TileResolution,
   zoomLevel: number,
-  units: 'meters' | 'degrees' = 'degrees'
+  // Default mirrors the maps-api twin ('meters'); callers slicing lon/lat
+  // geojson pass 'degrees' explicitly.
+  units: 'meters' | 'degrees' = 'meters'
 ): number {
   const circumference = units === 'meters' ? EARTH_CIRCUMFERENCE_METERS : 360;
   const tileWidth = circumference / 2 ** zoomLevel;
@@ -99,7 +106,10 @@ export function getMaxFeaturesByResolution(
  * it opted in via `fullTiles` on the source.)
  */
 export function isFullSourceTilejson(tilejson: TilejsonResult): boolean {
-  return tilejson.maxzoom === 0 && Boolean(tilejson.tiles?.[0]?.includes('full=true'));
+  return (
+    tilejson.maxzoom === 0 &&
+    Boolean(tilejson.tiles?.[0]?.includes('full=true'))
+  );
 }
 
 interface IndexedFeature {
@@ -142,12 +152,15 @@ export function buildFullTileIndex(features: Feature[]): FullTileIndex {
     let minY = Infinity;
     let maxX = -Infinity;
     let maxY = -Infinity;
-    eachPosition((feature.geometry as {coordinates: unknown}).coordinates, ([lng, lat]) => {
-      if (lng < minX) minX = lng;
-      if (lat < minY) minY = lat;
-      if (lng > maxX) maxX = lng;
-      if (lat > maxY) maxY = lat;
-    });
+    eachPosition(
+      (feature.geometry as {coordinates: unknown}).coordinates,
+      ([lng, lat]) => {
+        if (lng < minX) minX = lng;
+        if (lat < minY) minY = lat;
+        if (lng > maxX) maxX = lng;
+        if (lat > maxY) maxY = lat;
+      }
+    );
     if (!Number.isFinite(minX)) continue;
     const isPolygon = feature.geometry.type.includes('Polygon');
     const w = maxX - minX;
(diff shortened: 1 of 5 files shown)


jatorre:
## Review — approve, with small follow-ups (none blocking)

Reviewed the slicer math against the maps-api dynamic tiler (cloud-native #25467), built, linted, ran the suite, and exercised the path end-to-end in a live A/B harness (z10: one z0 fetch → 12 local slices, 0 per-tile fetches; renders identical to the per-tile control).

**Parity verified.** The points grid is a faithful port of the server SQL: same π/4 and π/360 constants, same `40075016.685…` circumference (server literal vs `2π·6378137`), and it correctly uses `2 ** aggLevel` rather than the 32-bit-unsafe `1 << aggLevel` (server runs it in SQL bigint; at full-mode depth `aggLevel` reaches ~28, which `1 << n` would overflow in JS). `getPointsAggregationLevel` is applied at the real tile `z`, reproducing the per-tile reduction. Agg-level (8), polygon cap (5000), lines cap (10000) and the per-resolution corrections all match the server defaults.

### Follow-ups (suggest addressing in review or a fast-follow)

1. **[Medium] Trim the public API surface.** `tileToBBox`, `getToleranceForSimplify`, `getPointsAggregationLevel`, `getMaxFeaturesByResolution` are exported from the package entry but no consumer imports them (deck.gl #45 only needs `isFullSourceTilejson` / `buildFullTileIndex` / `sliceFullTile`); they're only referenced by the tests. For a published lib this commits us to maintaining internal reduction math as public API. Prefer exporting just the three entry points (+ types) and importing the helpers from the relative module path in the test.
2. **[Medium] Document the server-settings coupling.** `DYNAMIC_TILES_POINTS_AGGREGATION_LEVEL` (8), `DYNAMIC_TILES_MAX_POLYGON_FEATURES` (5000), `DYNAMIC_TILES_MAX_LINES_FEATURES` (10000) are *settings* server-side (`MAPS_API_V3_*`). Defaults match today; if an operator overrides them, locally-sliced tiles silently drift from server tiles. A code comment naming the coupling would make a future settings change traceable.
3. **[Low] `getToleranceForSimplify` default units differ from the server twin** (`'degrees'` here vs `'meters'` server-side). Internally consistent (the cull works in degrees), but a footgun if the helper stays exported — another reason for #1.
4. **[Low] Test gaps:** no `lines` geomType test (the `minSize = pixel` branch + 10000 cap are unexercised), no direct sub-pixel-cull assertion (both polygon tests use features large enough to survive the cull), and `buildFullTileIndex` skip-paths (GeometryCollection / null geometry / Multi* recursion) are untested.
5. **[nit] `getMaxFeaturesByResolution('points', …)`** returns the polygon base via the ternary fallback — harmless (points don't use it) but surprising for an exported fn.

Marking ready so deck.gl #45 can un-draft once this publishes the slicer to `@carto/api-client`.

jatorre:
Addressed the review in `5ea0171`:

- **CI fix** — Prettier-formatted the two files (the `format:check` failure on Node 18/20/22; now green).
- **#1** — trimmed the public surface to `isFullSourceTilejson` / `buildFullTileIndex` / `sliceFullTile` (+ types); the reduction helpers are internal, tested via direct module import.
- **#2** — documented the maps-api settings coupling in a code comment.
- **#3** — `getToleranceForSimplify` now defaults to `meters` (mirrors the server twin); the lon/lat caller passes `degrees` explicitly.
- **#4** — added tests for lines sub-pixel cull, polygon sub-pixel cull, and `buildFullTileIndex` null/GeometryCollection skip + Multi* bbox span.

Full suite: 463 passed / 2 skipped, no type errors.

src/sources/client-side-tiler.ts
```diff
@@ -0,0 +1,264 @@
...
+const MAX_FEATURES_MULTIPLIER: Record<TileResolution, number> = {
+  0.25: 0.5,
+  0.5: 1,
+  1: 2,
+  2: 3,
+  4: 3,
+};
+
+/** Web-Mercator tile bbox in degrees: [west, south, east, north]. */
+export function tileToBBox(
```
zbigg, line 50: we already have `quadbin` in deps, can't we use   cellToBoundary(tileToCell(x,yz)) instead of crafting new math?

src/sources/client-side-tiler.ts
```diff
@@ -0,0 +1,264 @@
...
+  units: 'meters' | 'degrees' = 'meters'
+): number {
+  const circumference = units === 'meters' ? EARTH_CIRCUMFERENCE_METERS : 360;
+  const tileWidth = circumference / 2 ** zoomLevel;
+  const unitsPerPixel = tileWidth / (1024 * tileResolution);
+  return unitsPerPixel / 2;
+}
+
+/** Points aggregation grid level for a zoom (server parity). */
+export function getPointsAggregationLevel(
```
zbigg, line 80: please export this, we have use for it in builder too

src/sources/client-side-tiler.ts
```diff
@@ -0,0 +1,264 @@
...
+  if (typeof coords[0] === 'number') {
+    fn(coords as Position);
+    return;
+  }
+  for (const c of coords) eachPosition(c, fn);
+}
+
+/**
+ * Index the full source's features once (per loaded z0 tile). O(n) over
+ * coordinates; the result is reused for every subsequent local tile slice.
```
zbigg, line 143: can we/should we use established spatial index library like [rbush](https://github.com/mourner/rbush) (?) it has [O performance](https://stackoverflow.com/questions/48815868) of around "n + logM n" 

i know that it's tech preview now, but i believe we targed at dozens of thousands features so it's quite a work for JS without worker

zbigg (commented):
looks good, fea improvements proposed

src/sources/client-side-tiler.ts
zbigg: ok, actual library to use would be https://github.com/mourner/flatbush, it's faster, more suited 

nevertheless, expected improvements are relatively small so i believe we can leave these improvements for future, here's analysis https://gist.github.com/zbigg/7f97b1d6f34e8062ec05e1c8341a4bd6

src/sources/client-side-tiler.ts
zbigg: ok, don't export, just use getPointsAggregationLevel from https://github.com/CartoDB/carto-api-client/pull/304

